PR #31, opened 2018-11-16
Ignoring swp files

chrilves:


propensive:
Great, thank you for the first PR! Would you mind deleting the `.swp` file I accidentally checked in too? ;)

propensive:
@martinpallmann got there first, but I've merged both!

PR #39, opened 2018-11-20
fixes for repository cloning

propensive:
This changes the way repositories are cloned and checked out by Fury, and should fix the last blocker before releasing the Fury preview.


martinpallmann:
🎉

PR #40, opened 2018-11-22
Removed unnecessary imports in a variety of modules

propensive:


johnynek (commented):
How did you find these? Can we easily set to error on unused dependencies with fury?

(Warn on unused then error on warn).

propensive:
I added the `-Ywarn-unused:imports` flag, and `-Xfatal-warnings` flags to `scalac`, in all but one of the modules. I'm intending to maintain these going forward, though I need to do a bit more work elsewhere to fix an exhaustivity warning in the one remaining module.

I did all this using my Bloop build, but with Fury it would be a matter of setting,
```
fury param add -P Ywarn-unused:imports
fury param add -P Xfatal-warnings
```

I've got plans to experiment with how and where parameters are specified. Currently they're set independently for every module (which could be tedious with things you want to do globally or temporarily), but there are a few use cases to consider.

PR #46, opened 2018-11-28
Included very basic Dockerfile, which should trigger GCP build

propensive:


propensive:
This works, but I'm not going to merge it into this repository. I'll create a different repo for building Fury.

PR #70, opened 2019-01-08
Test ogdl

mzarnowski:
This pull request:
1. extracts OGDL Parser and added some basic test for it
2. extracts other - non-trivial - classes from ogdl.scala, since the file was getting very big and therefore not easy to read for someone unfamiliar with these concepts


propensive (changes requested):
Can you move all OGDL stuff into a new `ogdl` package, and filesystem-related stuff into an `io` package?

Other than that, the tests look like a good start!

propensive (approved):

PR #76, opened 2019-01-11
This should now correctly allow local sources to be defined

propensive:


tpasternak:
Integration tests fail:

    REPO  PATH           
    -     src/hello_world
    Fetching Git repository https://github.com/propensive/base.git.
    Checking out all sources.
    The schema 2.12.6 was not found.
    Fury daemon has shut down.

PR #78, opened 2019-01-12
Created a configuration for code formatting, and applied it

propensive:
This significantly increases the line-count for Fury, but it's a standard, at least.

This should be applied to all PRs before they're submitted. You will need scala-fmt 2.x.


src/build/build.scala
```diff
@@ -26,352 +26,387 @@ object ConfigCli {
...
-      description   <- io(DescriptionArg)
-      alias         <- ~Alias(aliasArg, description, optSchemaArg, moduleRef)
-      layer         <- Lenses.updateSchemas(None, layer, true) { s =>
-                           Lenses.layer.aliases } (_(_) += alias)
-      _             <- ~io.save(layer, layout.furyConfig)
+      cli <- cli.hint(SchemaArg, layer.schemas)
+      optSchemaArg <- ~cli.peek(SchemaArg)
+      cli <- cli.hint(AliasArg)
+      cli <- cli.hint(DescriptionArg)
+      optDefaultSchema <- ~optSchemaArg
```
ArturGajowy, line 95: May I suggest `align = most` so that those arrows (and many other things) are aligned neatly? Not that it matters (much) to me, but thought you might want to retain what's on the left side of the diff :)

ArturGajowy: It might be that overriding sub-properties of `align` won't work with `align = most`, disabling one or the others. That, at least, has been the case in 1.6

https://github.com/scalameta/scalafmt/issues/1202#issuecomment-393289730

propensive: Yes, I completely agree... I'll merge this and update to `align = most`.

PR #137, opened 2019-01-16
do not start fury if already started

### LINKED ISSUE #121: `fury start` is not always aware that it has started successfully
After installing as instructed, I do the following
```
artur@y510p:~$ fury start
Fury daemon is starting............................
Starting Fury is taking longer than expected.

# ^^^ (!)
# yet, Fury seems to be up and running:

artur@y510p:~$ fury about
The project has not been specified.

# to make sure no deamon would result in different behavior:

artur@y510p:~$ fury stop
Fury daemon has shut down.
artur@y510p:~$ fury about
connect: Connection refused
```

mzarnowski:
Also, some cleanup of fury script
fixes #121

### DIFF
--- a/etc/fury
+++ b/etc/fury
@@ -58,10 +58,10 @@ case "$1" in
     stopFury && startFury
     ;;
   "prompt")
-    fury "$@"  || printf " %{\e[0;30m%}[ ]"
+    fury "$@" 2> /dev/null || printf " %{\e[0;30m%}[ ]"
     ;;
   "completion")
-    fury "$@"  || printf "1:Command:((start:'start the Fury daemon'))\n"
+    fury "$@" 2> /dev/null || printf "1:Command:((start:'start the Fury daemon'))\n"
     ;;
   *)
     fury "$@"
--- a/etc/fury
+++ b/etc/fury
@@ -6,6 +6,10 @@ PORT="8462"
 CLASSPATH="$FURYHOME/lib/*"
 MAIN="com.facebook.nailgun.NGServer"
 
+nailgun(){
+  "${FURYHOME}"/bin/ng --nailgun-port "$PORT" "$@"
+}
+
 startFury() {
   SCALA=$(coursier fetch --classpath org.scala-lang:scala-reflect:2.12.8 com.facebook:nailgun-server:1.0.0)
   nohup java -cp "$SCALA:$CLASSPATH" "$MAIN" "$PORT" > /dev/null 2>&1 &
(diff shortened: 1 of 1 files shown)


tpasternak (approved):

etc/fury
```diff
@@ -1,7 +1,9 @@
 #!/bin/sh
 
+FURYHOME=$(dirname "$0")/..
```
propensive, line 3: So we can remove setting `FURYHOME` from the installation, can't we? Great!

etc/fury
```diff
@@ -1,7 +1,9 @@
...
+
 PORT="8462"
-CLASSPATH="$FURYHOME/lib/fury.jar"
+CLASSPATH="$FURYHOME/lib/*"
```
propensive, line 6: I didn't realize this worked...

propensive (approved):

etc/fury
```diff
@@ -51,14 +55,13 @@ case "$1" in
     stopFury && startFury && printf 'Fury daemon has been restarted.\n'
     ;;
   "prompt")
-    nailgun "fury.Main" -- "$@" 2> /dev/null || printf " %{\e[0;30m%}[ ]"
+    fury "$@"  || printf " %{\e[0;30m%}[ ]"
     ;;
   "completion")
-    nailgun "fury.Main" -- "$@" 2> /dev/null || \
-      printf "1:Command:((start:'start the Fury daemon'))\n"
+    fury "$@"  || printf "1:Command:((start:'start the Fury daemon'))\n"
```
propensive: Shouldn't we not still send `2> /dev/null`? If there's any error output, it shouldn't be visible to the completion system (which is the only reason this command should be called).

propensive: Also for the `prompt` line.

etc/fury
```diff
@@ -18,17 +18,18 @@ fury(){
   nailgun "fury.Main" -- "$@"
 }
 
+isRunning(){
+  fury help > /dev/null 2> /dev/null
```
propensive: I think `fury about` would be a better choice, because it does less work, but it's probably negligible. But could we do this even faster? I think this means invoking the python script *twice* now, and it was already noticeably slower than the native version. It's not by any means "slow", but it's not "instantaneous".

etc/fury
mzarnowski: But `*.jar` does not... ;)
A quirk of the jvm...

etc/fury
mzarnowski: I am not sure - it is used to install completions? But it probably does not have to be exported to the user environment? Let's think about it a bit more.

etc/fury
propensive: And it won't work if you're not careful to stop the shell trying to expand that `*`!

propensive (approved):

PR #145, opened 2019-01-17
Clean fury workspace

### LINKED ISSUE #68: provide a `clean` command
Fury creates various temporary files inside the `.fury` directory, which can be safely deleted (and Fury will recreate them), but this should be actionable through a command. The command should include options to delete:
- bloop configuration files
- repositories
- sources (sparse repos)
- bloop analysis files
- compiled classes

By default, this command should delete anything that's not used, and may be left over from previous configurations. But there should be a `--everything` option (or similar) which deletes everything, forcing it to be recreated as necessary.

mzarnowski:
closes #68

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -165,29 +165,6 @@ object BuildCli {
     } yield io.await()
   }
 
-  def clean(ctx: MenuContext) = {
-    import ctx._
-    for {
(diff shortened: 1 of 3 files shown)


propensive (commented):
Really minor points: Can you remove the `fury build clean` command, and change "clean sources" to "clean source checkouts" so nobody thinks that it's going to delete their source code. ;)

PR #170, opened 2019-01-22
prevent duplicate imports in a schema

### LINKED ISSUE #169: prevent duplicate imports

mzarnowski:
closes #169


propensive (approved):
Looks good. At some point it would be nice to review the various different collection types being used in the various algorithms, and check whether the performance could be improved. Currently, though, performance is not a problem.

PR #172, opened 2019-01-22
Include the current module's classfiles in the saved JAR

propensive:


tpasternak:
Confirmed, it works now

tpasternak (approved):
Tested manually - it works

propensive:
Thank you! The testing was probably more work than the implementation, I think...

PR #184, opened 2019-01-23
Mkdir before write to fury.log

### LINKED ISSUE #177: log directory is not created automatically
When Fury tries to write log files, it expects to find a `logs` directory inside `.fury`, but it isn't created automatically.

tpasternak:
closes #177


propensive (approved):
Thank you! I actually have this change in a slightly larger cleanup of `Layout`, but everything will get cleaned up during the merge. :)

PR #191, opened 2019-01-24
Use bundled coursier

### LINKED ISSUE #186: use bundled `coursier` from within Fury
As reported by Jakob Kozłowski, Fury tries to use the `coursier` that's on the PATH, but it doesn't always find this. It should use the one that's in the `$FURYHOME/bin` directory.

mzarnowski:
closes #186


propensive (approved):

propensive:
Looks good, thank you!

PR #203, opened 2019-01-25
verify modules before compiling or publishing

### LINKED ISSUE #26: verification of module references in builds
It's possible to create links in builds to modules which do not resolve. This is fine during setup of the build (for example, you can specify that the compiler to use should be `scala/compiler` before you've imported a workspace containing the Scala compiler. But we need to check that everything resolves before attempting to run a build or publish a workspace.

References to modules which do not resolve should be marked with a bright red `!!` (or similar) until they are fixed.

mzarnowski:
closes #26

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -18,6 +18,7 @@ package fury
 import fury.Args._
 import fury.io._
 import fury.error._
+import fury.module.Verifier
 import guillotine._
 
 import scala.concurrent._
@@ -195,6 +196,7 @@ object BuildCli {
       module      <- optModule.ascribe(UnspecifiedModule())
       hierarchy   <- schema.hierarchy(io, layout.pwd, layout)
       universe    <- hierarchy.universe
+      _           <- Verifier.verifyModule(universe, module)
       artifact    <- universe.artifact(io, module.ref(project), layout)
       artifacts   <- universe.transitiveDependencies(io, module.ref(project), layout)
       _           <- Bloop.server(layout.shell, io)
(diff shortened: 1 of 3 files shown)


src/build/build.scala
```diff
@@ -390,6 +392,9 @@ object LayerCli {
       cli           <- cli.hint(TagArg, suggestedTags)
       invoc         <- cli.read()
       io            <- invoc.io()
+      hierarchy     <- schema.hierarchy(io, layout.pwd, layout)
+      universe      <- hierarchy.universe
+      _             <- Verifier.verifyUniverse(universe)
```
propensive: This isn't quite necessary. We only need to verify all the modules in all the projects in all the schemas in the current layer. We don't need to verify the entire universe. But we do need the universe to check that the references resolve.

src/core/fury/module/Verifier.scala
```diff
@@ -0,0 +1,64 @@
...
+import fury.error._
+import fury._
+
+import scala.util.{Failure, Success}
+
+object Verifier {
+
+  def verifyUniverse(universe: Universe): Outcome[Set[ModuleId]] = {
+    val definedModules = for {
+      schema  <- universe.schemas.values
```
propensive: So this should be just `layer.schemas`.

src/core/fury/module/Verifier.scala
```diff
@@ -0,0 +1,64 @@
...
+      project <- schema.projects
+      module  <- project.modules
+    } yield module
+
+    definedModules.foldLeft(~Set[ModuleId]()) {
```
propensive: This fold is fine, but a recursive definition might be easier to work with. No need to change it.

src/core/fury/module/Verifier.scala
```diff
@@ -0,0 +1,64 @@
...
+      case (Success(verifiedModules), module) =>
+        verifyModule(universe, module, verifiedModules)
+      case (f, _) => f
+    }
+  }
+
+  def verifyModule(universe: Universe, module: Module): Outcome[Set[ModuleId]] =
```
propensive: I think this is more complicated than it needs to be. Don't worry - that's a good thing. Here's some pseudocode for what we need to do
```
for {
  schema <- layer.schemas
  project <- schema.projects
  module <- project.modules
  dep <- module.dependencies + module.compiler
 } yield universe.projects(dep.projectId).contains(dep.moduleId)
```

Then make this into a fold or a recursive method.

propensive (changes requested):

odisseus:
This pull request has become unmergeable.

Issue #208, opened 2019-01-25
underscore should not be allowed in module names


tpasternak:
You can add a module with underscore:

    fury module add -n my_module

But then you cannot add dependency to it from other module

    fury dependency add -M my_project/my_module

propensive:
Good catch! I think that the underscore should not be valid in a module name.

tpasternak:
@propensive isn't it done?

propensive:
@tpasternak It wasn't merged until now, due to the issue with Kaleidoscope, but I found a workaround.

PR #218, opened 2019-01-28
Tightening up of regular expressions

propensive:
This should fix issue #208, and hopefully make some of the checks a bit tighter in general, but probably doesn't fix everything, yet.


propensive:
This needs https://github.com/propensive/kaleidoscope/issues/17 to be fixed.

PR #221, opened 2019-01-29
Find dependency cycles before compilation

### LINKED ISSUE #29: check for cycles in dependencies
There's nothing to stop users creating cycles between module dependencies. I've not tested this, but it probably creates a stack-overflow error.

tpasternak:
closes #29 

- Check for cycles in dependency graph before compilation
- Graph algorithms moved to separate module


odisseus:
See #562.

PR #224, opened 2019-01-31
Support multiple undos

### LINKED ISSUE #21: increase the size of the undo buffer
It currently only has one level, so you can only undo once. A minimum of about five seems reasonable.

### LINKED ISSUE #128: `fury undo` doesn't seem to work + incorrect hint
```
➜  shapeless git:(master) ✗ fury undo
Could not find the file /home/plipski/shapeless/.layer.fury.bak. Run `fury init` to create a new layer.
```

Generally doesn't work and also suggests `fury init` rather `fury layer init`.

mzarnowski:
closes #21 
closes #128

It brings a new directory: `.fury/history` which stores the revisions of fury (`1.bak`, `2.bak`, etc.)
This revisions are 
a) created whenever the layer is modified
b) restored on `fury undo`

When there is nothing to `undo`, an error message appears: `No earlier revision can be found`

Currently, the limit of retained revisions is set to 16 - later it can (should?) be made configurable


test/undo_command/run.sh
```diff
@@ -0,0 +1,28 @@
...
+
+source $TESTROOTDIR/tests_lib
+
+start_bloop
+fury start
+fury layer init
+fury project add -n webpage
+fury module add -n hello_world
+fury source add -d wrong_src
+fury undo
```
tpasternak, line 13: Undo is not tested here. If `fury undo` does nothing, the test will still pass

mzarnowski: There is an invalid file inside - it will cause the compilation to fail

propensive (commented):
Great, thank you for this!

The change isn't a significant problem, but there was a little safety-net which has been removed by the change moving writing the file to disk out of the `Io` type and into `Layer`. Previously, it was only possible to write, if you had an instance of `Io`. And you could only get an instance of `Io` if you're not doing tab-completions. So it meant that it was never possible to write to disk when doing tab-completions (in the same way that it's impossible to print anything to STDOUT without an `Io`). Anyway, there were always a million ways to circumvent this, so it might get changed back. But that was the reasoning for it being written that way in the first place.

PR #240, opened 2019-02-08
Portable sed command in Makefile

jastice:
The sed command used in dist/bundle/bin/ng does not work on macOS (BSD) sed. Replaced with a portable version that should work both on Mac and Linux.


propensive:
Thanks, @jastice! Can you sign the CLA, please? :)

jastice:
done!

PR #249, opened 2019-02-12
Add a `make install` target

propensive:
You can now install the latest version of `Fury` in a single step - not very exciting!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,8 +149,10 @@ clean: clean-dist
 	rm -rf bootstrap/bin/fury
 	rm -rf bootstrap
 
-.PHONY: all publish compile watch bloop-clean clean-compile clean-dist clean test ci clean-ci test-isolated integration-isolated integration $(TESTS) all-jars
-
-.PHONY: download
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -149,8 +149,10 @@ clean: clean-dist
...
 	rm -rf bootstrap
 
-.PHONY: all publish compile watch bloop-clean clean-compile clean-dist clean test ci clean-ci test-isolated integration-isolated integration $(TESTS) all-jars
-
-.PHONY: download
 download: $(REPOS) dist/bundle/bin/coursier dist/bundle/bin/ng dist/bundle/bin/launcher dist/bundle/lib/$(NAILGUNJAR) bootstrap/scala
 	dist/bundle/bin/launcher --skip-bsp-connection $(BLOOPVERSION) # to download bloop
+
+install: dist/install.sh
+	yes | dist/install.sh
```
tpasternak: I don't like this approach.

1. It seems dangerous
2. Anyone can run `yes | make install` if needed 
3. Some apps may ask other questions then y/n. I think bloop asks for the target directory if it is going to be installed, so it'll end up in `$PWD/y`

propensive: Yes, good point. `make install` is actually a thing people commonly run, but it would still be interactive when it needs to be. I'll remove the `yes`.

PR #250, opened 2019-02-12
Remove single quotes from build/native menu description

propensive:
They were not being escaped, so they were causing completion to break. A
better solution would be to fix the escaping, but we're probably going
to rewrite it soon.


propensive:
The CI failure looks like a Bloop bug.

PR #251, opened 2019-02-13
Rename `--dependency`/`-M` to `--link`/`-l`

propensive:
While `--dependency` is maybe more informative, `--link` is conveniently
shorter, and begins with a different letter which isn't used elsewhere
(except for `--license`) so we can use a lower-case letter. It's a
slight improvement, I think.


propensive:
Incidentally, I think I already committed this change once as part of a larger PR, which means that it either got lost in a merge, or isn't merged yet...

tpasternak (approved):

PR #259, opened 2019-02-19
Centralize storage of repos and sources in .furyrc

propensive:


propensive:
The CI build was failing, but only because of the bug in Bloop.

PR #262, opened 2019-02-20
Introduce simplified testing infrastructure

propensive:
This should make writing tests a little easier. I have introduced two new directories inside `test`: `passing` and `pending`. The `passing` directory should contain tests we need to pass. The `pending` directory should contain tests which we want to pass at some time in the future, but currently don't (otherwise we would move them to `passing`).

The scripts have been moved to `etc/runtest` which takes two parameters: the directory to find the test in, and the name of the test, for example,
```
etc/runtest test/passing undo_command
```

The `etc/testall` script runs all the tests, and fails if any of the `passing` tests fail, but any failing tests in `pending` will not affect the exit status of `etc/testall`.

The test `repos_pull` was moved to `pending` because its output is currently unstable. We should rewrite it to produce the same output every time.


propensive:
Additionally, this pull request introduces the concept of `Reporter`s, which determine how compilation output gets presented to the user. You can choose how it's displayed using,
```
fury build compile --output linear
```
where the options to `--output` are `linear`, `graph`, or `quiet`.

propensive:
The current status of this PR is that integration tests appear to succeed, but don't actually run. I will investigate.

PR #272, opened 2019-02-26
Redesign cli

mzarnowski:


src/cli/fury/cli/Action.scala
```diff
@@ -0,0 +1,35 @@
...
+
+private[cli] object Action {
+
+  def apply[A](value: A): Action[Nothing, A] = new Completed(value)
+
+  /**
+    * Provides type-safe access for [[fury.cli.Parameter]] values (a parameter can only be accessed
+    * if it was specified as an action parameter - @see [[fury.cli.Action]]).
+    */
+  final class Bindings[+A](arguments: Arguments) {
```
mzarnowski, line 27: I am not happy with the separation of bindings and arguments, but during parameter completion we do not know which parameters were used when defining an `Action`, therefore we cannot be so restrictive.

Maybe there is another approach, but what we have to remember that :
1. completions are contextual (meaning they require the knowledge of layer -for example, to get the schema names),
2. completions are shared between actions - e.g. Schema or Project completions

src/cli/fury/cli/Parameter.scala
```diff
@@ -0,0 +1,103 @@
...
+
+private[cli] case class Flag(declaration: Declaration) extends Parameter[Boolean] {
+  override def apply(arguments: Arguments): Try[Boolean] =
+    arguments.valueOf(this) match {
+      case Some(_) => Try(true)
+      case None    => Try(false)
+    }
+}
+
+// TODO - it is "named" (as in "--name value"), but what else can it be called?
```
mzarnowski, line 30: Any suggestion would be appreciated

src/cli/fury/cli/Parameter.scala
```diff
@@ -0,0 +1,103 @@
...
+    ): Parameter[Option[A]] = {
+    val declaration = createDeclaration(longName, shortName, description)
+    Named(declaration)(extractor.andThen(Some.apply), Some(Success(None)))
+  }
+
+  case class InvalidParameter(parameter: Parameter[_], value: String = "") extends Exception
+
+  /**
+    * name: someName
+    * aliases: (--someName, -s) TODO find better name
```
mzarnowski, line 96: Any suggestion would be appreciated

mzarnowski (commented):
There are some things I am not sure, so please - share your thoughts!

propensive:
Hi @marek1840, I won't get a chance to look at this for another couple of days, but when I do, I'll make sure I look at it very thoroughly. I'd like to spend some time experimenting with it.

PR #273, opened 2019-02-27
Introduce cli completion

mzarnowski:
builds on top of #272 - starts from c267ac1

### DIFF
--- a/src/cli-completions/fury/cli/CompletionCommand.scala
+++ b/src/cli-completions/fury/cli/CompletionCommand.scala
@@ -1,0 +1,26 @@
+package fury.cli
+
+import fury.cli.CompletionDispatcher.complete
+import fury.cli.CompletionParameters._
+import fury.cli.Parameter.parameter
+import fury.cli.Shell.Zsh
+import fury.cli.CompletionFormatter._
+
+final private[cli] class CompletionCommand(interface: CLI[_], sources: ParameterValueSources)
+    extends Action(ShellParameter :+ OffsetParameter)(bindings =>
+      for {
+        shell  <- bindings(ShellParameter)
+        offset <- bindings(OffsetParameter)
+
+        argumentPosition = shell match {
+          case Zsh => offset - 1
+          case _   => offset
+        }
+
+        suggestions   <- complete(interface, argumentPosition, sources, bindings.positional)
+      } yield format(suggestions)(shell, offset))
+
+private[cli] object CompletionParameters {
+  val ShellParameter  = parameter("shell", "s", "shell (e.g. bash, zsh)")(Shell.parse)
+  val OffsetParameter = parameter("offset", "o", "offset of a command")(s => Some(s.toInt))
+}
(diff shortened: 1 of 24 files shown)


src/cli-completions/fury/cli/ParameterValueSource.scala
```diff
@@ -0,0 +1,20 @@
...
+
+trait ParameterValueSources {
+  def sourceFor(parameter: Parameter[_]): Option[ParameterValueSource]
+}
+
+trait ParameterValueSource {
+  final def values(prefix: String)(arguments: Arguments): Seq[ParameterValue] =
+    matchingValues(prefix)(arguments) match {
+      case Success(values) => values.filter(matches(prefix))
+      case Failure(_)      => Nil // TODO log the cause
```
mzarnowski, line 13: once we get the logging done, it would be nice to know the cause

src/cli-completions/fury/cli/ParameterValueSource.scala
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+  protected def matchingValues(prefix: String)(arguments: Arguments): Try[Seq[ParameterValue]]
```
mzarnowski, line 16: better name - anyone?

src/cli-completions/fury/cli/ShellCompletion.scala
```diff
@@ -0,0 +1,19 @@
+package fury.cli
+
+sealed private[cli] trait ShellCompletion
+
+private[cli] object ShellCompletion {
+  case class MenuCompletion(menu: Menu[_], commands: Seq[CLI[_]]) extends ShellCompletion
+
+  case class ParameterNameCompletion(
```
mzarnowski, line 8: `ParameterNameCompletion` and `ParameterValueCompletion` could be potentially merged together, but since `bash` completions are much simpler than `zsh` we would still require to somehow pass the context of what is being currently completed 

in `zsh` it does not matter, because parameter completion looks always the same, while in `bash` these are either the parameter names ('--param1' '--param2' ...) or values for the parameter

src/cli-completions/fury/cli/CompletionSource.scala
```diff
@@ -0,0 +1,76 @@
...
+        case NameCompletion  => ParameterNameCompletion(name, parameter, values)
+        case ValueCompletion => ParameterValueCompletion(name, parameter, values)
+      }
+
+  private def matchingParameters(parameters: Parameters[_], namePrefix: String) =
+    for {
+      parameter <- parameters.all.toSeq
+      alias     <- parameter.aliases
+      if alias.startsWith(namePrefix)
+    } yield (alias, parameter)
```
mzarnowski, line 61: regrettably, we must filter parameter aliases ("-p" does not match "--param", even though it is the same parameter), hence the tuple

src/cli-completions/fury/cli/CompletionCommand.scala
```diff
@@ -0,0 +1,26 @@
...
+import fury.cli.CompletionFormatter._
+
+final private[cli] class CompletionCommand(interface: CLI[_], sources: ParameterValueSources)
+    extends Action(ShellParameter :+ OffsetParameter)(bindings =>
+      for {
+        shell  <- bindings(ShellParameter)
+        offset <- bindings(OffsetParameter)
+
+        argumentPosition = shell match {
+          case Zsh => offset - 1
```
mzarnowski, line 16: `zsh` seems to index the arguments from `1`, not `0`...

PR #279, opened 2019-03-05
Implemented recursive search for layer.fury till the filesystem boundary

### LINKED ISSUE #266: recursively search parent directories to find `layer.fury`, like `git`
If there is not a `layer.fury` file in the current directory, Fury should search up the tree to see if `fury` commands are being run from within a subdirectory of a Fury project, and choose the ancestor directory which contains `layer.fury` as the base.

odisseus:
This fixes #266.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -6,7 +6,7 @@ steps:
 
 - name: 'gcr.io/cloud-builders/docker'
   args: ['run', '-u', 'bash_user', '-w', '/home/bash_user', '-t', 'fury/ci', '/test/run_all']
-  timeout: 120s
+  timeout: 240s
   id: 'integration'
 
 - name: 'gcr.io/cloud-builders/docker'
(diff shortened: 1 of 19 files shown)


propensive (changes requested):
I think I would like to see only a single `Layout` instance visible in all of the for comprehensions, and I'd like `cli.layout` to be changed to always return the main layout, not the "inner" one. Instead of having just the value `pwd` inside of `Layout`, could we have `pwd` (which is the directory `fury` is executed from) and a new value called `base` which is the directory containing the file `layer.fury`?

odisseus:
What about the case when the main layout doesn't exist yet? E. g. running `fury layer init` in an empty directory.

propensive:
I can't check this now, because I'm having some problems with Bloop... it's refusing to compile anything.... I will take another look in the morning.

propensive:
Sorry this took me so long, @odisseus!

PR #295, opened 2019-03-13
IPFS-stored layers and support for editing nested layers

propensive:
This PR makes some significant changes to how layers are handled in Fury. I'll try to summarize the differences.

We now store layers in IPFS, instead of Git. This means that we can address them by a (long) hash of their content, so now, whenever we refer to a layer (for example from another layer) we do so by its hash. We no longer have to add a `repo` for a layer, and then import it: we simply refer to its hash.

So now, when you do anything to edit a layer, it modifies the layer, just as it did before, stores a serialized version of it in IPFS, and gets a new hash. We now only need to store the hash in the working directory (not the whole layer file), as a reference to the current version of the layer.

As before, we can have layers which depend on other layers (which depend on other layers). So your working layer might depend on a `typelevel` layer, which might depend on a `circe` layer. If we wanted to experiment with a change to Circe, we can now directly work on the "Circe" layer and have the changes reflected immediately in our layer. Firstly, we would run:
```
fury layer select -l ./typelevel/circe
```
This would update a pointer in a file in the current working directory to point to the `circe` layer, referenced from the `typelevel` layer. The current working directory would still point to the same "base" layer, though: we are just looking within it. But any other Fury commands now operate on the `circe` layer. We can change whatever details we like, before switching back to the base layer, and attempting recompilation again.

Because changes to the `circe` layer affect its hash, the reference to it from the `typelevel` layer must also change, and thus the hash of the `typelevel` layer will change, and so will the "base" layer. Every time a change is made to a nested layer, Fury will update the hash in the current working directory to point to the current version of the layer and all its nested layers.

IPFS also enables anyone else on the Internet to access a build by its hash, and if your computer is running an IPFS daemon, it will serve the file anywhere on the Internet using a hash. This will be how people will share builds.

IPFS hashes are immutable, but I'm also going to build a mutable name-based layer on top of them to allow people to publish builds under a known name. This hasn't been fully designed yet.


odisseus (commented):
I see a security issue with this concept.

Users _will_ publish sensitive information (personal data, private keys etc.) by accident. With a central authority like Github, you could contact them to scrub your data off their servers (force pushing really just hides the commit, but it is still accessible by hash), or, failing that, delete your whole repository. Does IPFS provide any way to make sure your data isn't actively distributed by some other node?

Also, does IPFS have anything similar to the .gitignore mechanism to make sure certain files and directories are not published?

propensive:
@odisseus It's an important point, but at the moment, the only information that gets stored in IPFS is the layer OGDL file; not source files... but that might change. The build may still contain IP, so it shouldn't be shared accidentally.

Anyhow, I'm thinking that in a commercial environment, that's not acceptable. So I'm probably going to offer a version of Fury that will only publish things to IPFS encrypted. But the security model is something I'm thinking very carefully about.

Essentially, with this scheme, there would need to be several failures all happening together for data to accidentally leak.

odisseus:
Another issue is tracking of changes between layer versions. Suppose our build depends on layer Foo with hash A, and we want to update that dependency to the latest version. Foo is also avaliable in hashes B and C, both with later release date. How do we know which one is the "best" version, what exactly has changed between them, and whether they are interchangeable with version A at all?

adamsmo:
There Is another issue: Owner of the project would like to publish updates to a library/project and share it without the need for releasing the exact hash with every single update of the project.
At the same time, there has to be a way for the users to verify if the build is from the project maintainer and not from a malicious third party.

I was thinking about solutions, and I found out that Fury can use:

GPG:
Use GPG infrastructure of trust and public key distribution if you trust public key you trust signed build.
There are existing clients for macOS Linux and windows, but none the less it is an additional tool that you have to install and interact with from Fury itself which may be difficult especially on Windows. GPG is also not as popular as ssh 

SSH key:
It is more popular than GPG, and Git uses it, even client for Windows uses ssh keys in the same format as on macOS/Linux. It is relatively simple to sign files with ssh
https://superuser.com/questions/308126/is-it-possible-to-sign-a-file-using-an-ssh-key
There is no additional software to install - we can use ssh key with crypto libraries.
The disadvantage is that there is no trust management; there is no public key distribution network. So there has to be some way to find out which public key belongs to who, there is also the issues of invalidating stolen/lost keys that are handled by GPG.

Https:
This solution provides a way to verify build file without encryption. The assumption is that every project has some page, like https://typelevel.org/cats/ or https://scalaz.github.io/scalaz-zio/ that is protected with https, so the way it can work is to ask project maintainer to publish a file named releases.txt under some path protected with https. The idea is that this file will have mappings that will look like this
`release_name -> IPFS hash`
When a user uses a library, she picks a name like 
`typelevel.org/cats % 1.0-M5`  that Fury resolves by fetching releases.txt with https
The solution requires no additional software, is platform independent; trust is managed by authorities issuing SSL certificates. Nowadays it is easy to get a free certificate from organizations like LetsEncrypt.

The maintainer is not required to use shes own domain she can relay on pages already protected with https like GitHub pages. Path, in this case, may like this 
`adamsmo.github.io/final-tagless % 1.0 `
with GitHub page setup like this https://adamsmo.github.io/final-tagless/releases.txt

Advantages: no signing, IPFS is doing all the work by ensuring file integrity with hash, no need to manage sensitive data (private keys), for implementation requires HTTP client that supports ssl

Disadvantages: longer names for the libraries `typelevel.org/cats` instead of `cats` but this can be solved with IPNS

WDYT?

PR #297, opened 2019-03-13
WIP: Bsp server support

jastice:
Current status: very basic bsp server support. Works well enough to import fury itself into IntelliJ latest Nightlies (protocol version 2.0.0-M4, partially incompatible with 2.0.0-M3). Compiling is not yet supported.

Adds 2 commands:
* `fury bsp init` - writes a bsp connection file to the current workspace
* ` fury bsp run` - starts a bsp server

Starting the bsp server via menu command didn't work in conjuction with IntelliJ, probably due to Nailgun, so the connection file for now defines a command to start a Fury JVM process directly.

The server is currently stateless, so it may recompute fury structure on every request, which may or may not have performance implications for larger projects.

Schemas aren't respected, it just uses the mainSchema

Overall this is pretty basic and kind of dirty and some code is duplicated so as not to mess to much with the rest of Fury, but I'd appreciate general feedback how to make it fit in better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ dist/bundle/bin/coursier: dist/bundle/bin/.dir
 	chmod +x $@
 
 jmh_jars=org.openjdk.jmh:jmh-core:1.21 org.openjdk.jmh:jmh-generator-bytecode:1.21 org.openjdk.jmh:jmh-generator-reflection:1.21 org.openjdk.jmh:jmh-generator-asm:1.21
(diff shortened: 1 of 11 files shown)


propensive (approved):
I haven't taken a good look through `bsp.scala`, but I like that it's self-contained... so I don't have to worry about how it interacts with anything else. I think this is good enough to be merged, in particular because it's something we can iterate on, and the easier we can make that, the better. Thank you, @jastice!

PR #305, opened 2019-03-14
Use msg interpolator to save some characters, "fix" core-test compilation

kubukoz:
Just a simple change to start getting familiar with the codebase.

Also: "fixing" tests by adding imports (I'm pretty suspicious about this, as the tests being broken would mean they are not being run on CI or the compilation errors are ignored - if there are some global imports configured, let me know how to run the tests with them).

### DIFF
--- a/src/core-test/fury/tables/TablesTest.scala
+++ b/src/core-test/fury/tables/TablesTest.scala
@@ -1,0 +1,25 @@
+package fury
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+import fury.core._
+import fury.strings._
+
+import probably._
+
+object TablesTest extends TestApp {
+  override def tests(): Unit = {
+    test("contextString with showSchema=true") {
+      Tables(Config())
+    }.assert { tables =>
+      tables.contextString(msg"foo", true, msg"bar", msg"baz").string(Theme.NoColor) == "foo/bar/baz"
+    }
+
+    test("contextString with showSchema=false") {
+      Tables(Config())
+    }.assert { tables =>
+      tables.contextString(msg"foo", false, msg"bar", msg"baz").string(Theme.NoColor) == "foo/baz"
+    }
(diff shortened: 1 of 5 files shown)


src/core-test/fury/tables/TablesTest.scala
```diff
@@ -0,0 +1,25 @@
...
+import fury.strings._
+
+import probably._
+
+object TablesTest extends TestApp {
+  override def tests(): Unit = {
+    test("contextString with showSchema=true") {
+      Tables(Config())
+    }.assert { tables =>
+      tables.contextString(msg"foo", true, msg"bar", msg"baz").string(Theme.NoColor) == "foo/bar/baz"
```
mzarnowski: Most of this line should probably ( ;)) be in a 'test' block like:
```
test(...) {
//given
val tables =??? 
//when
tables. Do something 
}. assert { //then
result => result == expected 
} 
```
(od course comments should not be included - they are here just to show intent) 
In my opinion this approach greatly increases readability

mzarnowski (changes requested):

src/core-test/fury/tables/TablesTest.scala
kubukoz: Thanks, sounds good (although it doesn't align with the style of other tests). I'll update accordingly.

src/core/tables.scala
```diff
@@ -53,14 +51,14 @@ case class Tables(config: Config) {
     case s          => Ansi.green("-" + s)
   }.join("\n")
 
-  implicit private def option[T: AnsiShow]: AnsiShow[Option[T]] = _ match {
+  implicit private def option[T: AnsiShow]: AnsiShow[Option[T]] = {
```
propensive, line 54: Nice. I didn't think of this!

propensive (approved):

Issue #280, opened 2019-03-06
use jgit instead of `git`


odisseus:
JGit can probably be used for that.

propensive:
We need to ensure that all of the `git` commands [defined here](https://github.com/propensive/fury/blob/master/src/core/shell.scala#L52-L175) can be implemented with `jgit`.

Possibly some of these are not used any more, and can be removed.

propensive:
It's been recommended to me that JGit is not reliable enough, so we're going to keep using the shell commands.

PR #307, opened 2019-03-14
Remove unused git code

kubukoz:
As mentioned in #280, some methods were possible to be removed since they weren't used.


propensive (approved):
This will help a lot with #280.

Issue #300, opened 2019-03-13
always generate bsp config


propensive:
I can't see any reason why we shouldn't create `.bsp/fury.json` when we initialize the repository. I don't think it needs to be a separate command, and it's a cheap operation. Any reason not to, @jastice?

jastice:
Generally agree. Reasons it might still need to be separate command:
* projects that have not yet created the bsp file. `fury layer init` currently overwrites an existing `layer.fury`, that might have to be changed
* generating a new `fury.json` when fury installation location changes. Eventually Fury should set up the bsp config in the user/system space anyway.

jastice:
I think there was still the issue that after cloning a repo, there was no way to generate a bsp config except modifying the build in some arbitrary way

PR #310, opened 2019-03-14
Always create `.bsp/fury.json` whenever `layer.fury` changes

propensive:


jastice:
I suppose this works but it seems hacky to have to change the layer to generate a bsp connection file :)

odisseus:
This PR is probably related to #300.

PR #312, opened 2019-03-15
Added tests for OGDL serialization.

odisseus:


src/ogdl-test/fury/ogdl/OgdlSerializerTest.scala
```diff
@@ -0,0 +1,72 @@
...
+        "A" ->
+          graph("B" -> graph("C", "D")),
+        "E"
+      ))
+    }.assert(_ == "A\tB\tCD\nE\n")
+  }
+
+  private def write(node: Ogdl): String = Ogdl.serialize(node)
+
+  implicit private[this] def buffer(string: String): ByteBuffer        = wrap(string.getBytes())
```
propensive, line 68: I was a bit uncomfortable about these implicits, and I'd still prefer them not to be there (even safely `private`), but it's better to have the tests, so I'll merge and they can be cleaned up later.

propensive (approved):

PR #319, opened 2019-03-19
Enabled resolving binaries from alternative repositories.

odisseus:


propensive (changes requested):
Could you change the `IllegalStateException` to one of Fury's exception types (or if you can't find a suitable one, create your own) that's defined in `exception.scala`? You will need to make sure it's handled appropriately in `recovery.scala`. Other than that, it all looks good, thank you!

odisseus:
I have refactored error handling and also added an integration test.

propensive (approved):
Looks great, thank you @odisseus!

Issue #324, opened 2019-03-25
The command `fury kill` fails silently


odisseus:
This tends to happen after Fury is rebuilt using the makefile.

odisseus:
This issue doesn't appear anymore in the latest versions. Additionally, `fury kill` now has an optional `--force` flag, which kills any process that looks like Fury.

Issue #317, opened 2019-03-18
Fury fails without any output


odisseus:
When any task fails for whatever reason, there should be an error message. Sometimes the tasks fail silently, which can be easily mistaken for a success.

Here's an example log of commands:
```
$ fury layer init
$ fury project add -n scalaj-http
$ fury module add -n core
$ fury project list
$ fury module list
```
At this point it is clear that something has gone wrong, but there are no error messages whatsoever. Restarting the daemon seems to help:
```
$ fury restart
Fury daemon has shut down
Fury daemon is starting....
Fury daemon is running
$ fury project list
The project has not been specified.
$ fury layer init
Created empty layer.fury
$ fury project add -n scalaj-http
Set current project to scalaj-http
$ fury module add -n core
Set current module to core
$ fury project list
~/devel/scalaj-http-fury
   PROJECT      MODULES  DESC  LICENSE  COMPILER
>  scalaj-http  ■              unknown  java    
$ fury module list
~/scalaj-http-fury/scalaj-http
   MODULE  TYPE     DEPENDENCIES  SRCS  BINS  COMPILER       TYPE     PARAMS
>  core    library                            java/compiler  library  
```

PR #326, opened 2019-03-26
Added a function that prints stderr after the command has completed.

odisseus:
This pull request makes stderr visible without disrupting the pretty print.
This change doesn't fix any problems by itself, but it brings to the surface the causes of issues like #317 and #324.


propensive (approved):
This reminds me how horrible bash scripting can be... Looks good. :)

Issue #48, opened 2018-11-30
publishing to Maven Central


propensive:
It should be possible to publish to Maven Central (or Bintray) from Fury, if a project has only binary dependencies which are already on Maven Central (or Bintray). [This file](https://github.com/xerial/sbt-sonatype/blob/master/src/main/scala/xerial/sbt/Sonatype.scala) should contain the essence of what's required to publish to Maven Central. I think we can probably make everything a lot simpler than it is in here, though.

If possible, I'd prefer not to have any dependencies on third-party libraries, so if we can implement HTTP calls using `java.net` functions, that would be preferable. I'd also like to use [Xylophone](https://github.com/propensive/xylophone/) for XML, though this project needs some maintenance first.

propensive:
We should use the code in Mill to do publishing to Maven Central / Sonatype:
https://github.com/lihaoyi/mill/tree/de175e69977082e35539097a54d381e465dddf8e/scalalib/src/publish

PR #330, opened 2019-03-28
Added limited support for artifact publishing. (WIP)

odisseus:
This pull request aims to resolve issue #48.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ VERSION=${shell sh -c 'cat .version 2> /dev/null || git --git-dir git/fury/.git 
 MKFILE := $(abspath $(lastword $(MAKEFILE_LIST)))
 ROOTDIR := $(dir $(MKFILE))
 BLOOPVERSION=1.3.2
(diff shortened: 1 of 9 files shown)


propensive (commented):
Thanks, @odisseus. This looks like a great start. I will try to spend some time to integrate this with a UI and some changes to the Fury model. Not ready to approve yet (I'm still preparing for my talk in a couple of hours), but it shows how the API might be used.

Do you think it would be possible for Shuttlecraft to give any progress updates while it's publishing? Notifications as each file is uploaded, for example?

odisseus:
Progress reports are [definitely](https://github.com/VirtusLab/shuttlecraft/issues/5) a good thing to have.  However, my top priority is implementing actual support for anything beyond publishing unsigned snapshots to a Nexus repository.

odisseus:
I have refactored Shuttlecraft to use scala-logging and logback, so the progress reports are looking better now, and are configurable (at least in theory, because at the moment there seems to be no way to package a logback.xml into the distribution of Fury).

PR #335, opened 2019-04-05
Better connection management

propensive:
This might help improve the concurrency issues with running Fury on many different projects. We now use an explicit thread pool for launcher connections, and explicitly cancel launcher futures when they're shutdown.


propensive:
CI succeeded in the previous commit, and the changes in this commit are negligible. The the failure appears to be the usual Bloop error, during unit tests, so I'm merging to save time...

PR #339, opened 2019-04-06
Added an option to kill Fury and Bloop processes for sure.

odisseus:
Fury tends to leave a lot of processes lingering in the background even after it is stopped, and these processes don't respect SIGTERM. Perhaps this is related to Fury being recompiled and restarted, but I'm not sure. Regardless, I think that an easy and sure way to nuke these processes will save quite some hassle during development.

### DIFF
--- a/etc/fury
+++ b/etc/fury
@@ -41,7 +41,22 @@ coursier(){
 }
 
 killFury() {
-  nailgun ng-stop > /dev/null && printf 'Fury daemon has shut down\n'
+  case "$1" in
+    "--force")
+      nuke &
+      ;;
+    *)
+      nailgun ng-stop > /dev/null && printf 'Fury daemon has shut down\n'
+      ;;
+  esac
+}
+
+nuke(){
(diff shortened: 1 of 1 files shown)


etc/fury
```diff
@@ -41,7 +41,23 @@ coursier(){
...
+      ;;
+  esac
+}
+
+nuke(){
+bash <<END
+  printf 'Killing everything that looks like fury or bloop...\n'
+  pkill --signal SIGKILL -f fury
+  pkill --signal SIGKILL -f bloop
+  pgrep -f fury || pgrep -f bloop && printf 'Processes with these PIDs have survived...\n' || printf 'Done! \n'
```
tkroman: should this be `pgrep -f fury bloop`? otherwise you'll only see first command's output

odisseus: Good point! Fixed.

propensive (approved):
I think this should be a temporary option until we no longer need it. :)

PR #342, opened 2019-04-11
Replaced the hashes in an error message with module references.

odisseus:


propensive (approved):

propensive:
Approved, but awaiting a successful build...

PR #343, opened 2019-04-11
Shuffled `fury bsp` command a little more.

odisseus:


propensive (changes requested):
Can you change `which fury` to `command -v fury`? The latter is POSIX-compliant.

propensive (changes requested):
Can you change `which fury` to `command -v fury`, so that it's POSIX-compliant?

propensive:
Ok, airport wifi *was* working after all. Now you have two reviews! ;)

odisseus:
Changes done

propensive (approved):

PR #364, opened 2019-05-07
Hierarchy of module dependencies in the makefile

odisseus:


odisseus:
Also made the dependent makefiles use the same version of Scala.

Makefile
```diff
@@ -70,21 +71,55 @@ dependency-jars: dist/bundle/bin/coursier dist/bundle/lib
...
+src/jsongen: src/strings
+	$(compile-module)
+
+src/io: src/strings
+	$(compile-module)
+
+src/ogdl: src/io
+	$(compile-module)
+
+src/fury-core: src/jsongen src/ogdl
```
propensive: Why is this called `src/fury-core` instead of `src/core`?

odisseus: Because it wouldn't compile. I haven't investigated this problem, but I think this is caused by name collision with "core" modules of various source dependencies.

_P.S._ It does compile; apparently the error I got the previous time I checked was caused by something else.

PR #366, opened 2019-05-13
Fixed the error message for conflicting projects.

odisseus:
This message used to say something like "Your dependency tree contains references to two or more conflicting projects: ProjectId(magnolia), ProjectId(contextual)", which doesn't give the slightest hint about which schemas actually cause the conflict.


propensive (changes requested):
Looks great! Do you think you could incorporate the "diamond" graphic in the output, that's currently implemented (but not yet used anywhere), here:
https://github.com/propensive/fury/blob/master/src/core/graph.scala

Issue #369, opened 2019-05-17
saving JAR files does not get classes from correct directories


propensive:
Since upgrading to the latest version of Bloop, saving JAR files does not work any more.

The reason is that, for the purposes of keeping different Bloop clients isolated, the output directories specified in the Bloop configuration files are not the exact directories that are used: Bloop will create new directories with the same name, plus a suffix. Class files will be put into these new directories.

So, when saving a JAR file, for each module, Fury will need to specify this rewritten directory name instead of the one it thinks it should be using.

In order to do this, I think Fury will need to issue a BSP request to get the classpath for each module, and then use this to work out the correct output directory corresponding to each module. I suggest that this should be done when constructing the `Compilation` object, and stored in that object. I've recently modified the code to construct `Compilation`s asynchronously whenever configuration files change, and then store them in a global hashmap. (They're constructed synchronously if they're not already in the hashmap.)

odisseus:
It keeps happening! Even with Bloop versions 1.3.0 and 1.3.2.

odisseus:
See https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#scalac-options-request

odisseus:
This issue has been fixed by pull requests #403 and #411.

PR #374, opened 2019-05-22
Include a new test to check whether JAR-saving is working

propensive:


odisseus:
This is a useful test to have, but I believe it won't pass once the issue #369 is fixed. In particular, the expected output should include names of the class files.

propensive:
Yes, that's fine. I'd like the development process to be that we fix the bug while using the test to check whether it's doing the right thing. And as part of the PR which fixes the code, we also check in fixes for the test, as necessary.

PR #386, opened 2019-06-02
Fix benchmarking

tpasternak:


tpasternak:
Short description of changes: 
1. `"asm"` -based generator does not work with java 8 and newer we can use `"reflection"` instead 
https://github.com/bazelbuild/rules_scala/issues/295

2. we send `Print(msg)` messages with benchmarking info after the streaming is closed. I don't know what's the right solution, this is just a quick fix.

PR #417, opened 2019-07-03
Implemented assembly of fat JARs

odisseus:
This is an alternative to the pull request #418.

### DIFF
--- a/src/core/args.scala
+++ b/src/core/args.scala
@@ -56,6 +56,7 @@ object Args {
   val DescriptionArg =
     CliParam[String]('D', 'description, "specify a brief description of the project")
   val ForceArg  = CliParam[Unit]('F', 'force, "force this operation")
+  val FatJarArg = CliParam[Unit]('F', Symbol("fat-jar"), "package the module along with all its dependencies")
   val FileArg   = CliParam[Path]('f', 'file, "destination file")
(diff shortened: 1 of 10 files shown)


src/core/args.scala
```diff
@@ -56,6 +56,7 @@ object Args {
   val DescriptionArg =
     CliParam[String]('D', 'description, "specify a brief description of the project")
   val ForceArg  = CliParam[Unit]('F', 'force, "force this operation")
+  val FatJarArg = CliParam[Unit]('F', Symbol("fat-jar"), "package the module along with all its dependencies")
```
propensive, line 59: I'm tempted to try to find a different choice of symbol than `F`. Using `--force` should overwrite an existing file, so we might need both. Suggestions:
`--recursive`
`--all`
`--deep`
Any other ideas?

propensive (approved):

src/core/args.scala
odisseus: Anything is ok with me. The current name of this argument probably isn't the best, but I had to choose some name for it, and I didn't have any better idea at the time. 

Fat JARs are also called [assembly](https://github.com/sbt/sbt-assembly) JARs, so maybe the argument to create such JAR could be `--assembly`.

PR #420, opened 2019-07-15
[WIP] Implemented a workaround for the quirks of BSD sed.

odisseus:
The BSD version of sed, also used in the Apple operating systems, has a different syntax for editing files in place and doesn't like Unicode characters.


propensive (approved):
Was the non-ASCII a problem somewhere?

PR #421, opened 2019-07-19
Fix Make build on Mac

jastice:
Fix Fury make build on mac

make build was broken through
1. non-portable sed command, incomatible with Mac/BSP sed
2. non-portable way of determining FURYHOME
3. non-portable definition of compile-module (it would simply not be executed, silently generating empty fury.jar)

Make is not a good build tool :(

Also fix generation of empty tmpfiles, and remove fury direct / change it to fury standalone


jastice:
presumably obsoletes #420

propensive (approved):

propensive:
You're right about `make`. What we need is a new build tool.

kubukoz:
JON NO

Issue #397, opened 2019-06-08
repeated compilation fails due to pipe errors


odisseus:
When compiling the same project several times in a row, subsequent compilations usually fail with error logs containing messages like "Pipe closed" or "Pipe broken". This is apparently a side effect of pull request #375.

odisseus:
It turns out that the `compile` command may create a BSP connection more than once. Sometimes only one of these connections throws an exception.

propensive:
I merged what should be a fix for this. It's difficult to test, as the broken-pipe errors weren't happening consistently, but let's reopen this if it turns out not to work.

odisseus:
The pipe errors still happen sometimes, especially during the cloud build:
```
----------------------------------------------------------------------------------------------------
 build-zio: Build ZIO from source
----------------------------------------------------------------------------------------------------
Cloning repository at https://github.com/propensive/.scala.git
Checking out * from repository scala
Resolving org.openjdk.jcstress:jcstress-core:0.4
Resolving org.specs2:specs2-core_2.12:4.7.1
Resolving org.specs2:specs2-matcher-extra_2.12:4.7.1
Resolving org.specs2:specs2-scalacheck_2.12:4.7.1
Resolving com.github.ghik:silencer-lib_2.12:1.4.2
Cloning repository at https://github.com/zio/zio.git
Checking out {core/jvm/src/main/java, core/jvm/src/main/scala, core/jvm/src/main/scala-2.12, core/shared/src/main/scala, core/shared/src/main/scala-2.12+} from repository zio
Checking out core/jvm/src/main/java from repository zio
Checking out {stacktracer/jvm/src/main/scala, stacktracer/shared/src/main/scala} from repository zio
Checking out {streams/jvm/src/main/scala, streams/shared/src/main/scala} from repository zio
Cloning repository at https://github.com/portable-scala/portable-scala-reflect.git
Checking out jvm/src/main/java from repository portable-scala-reflect
Checking out {test/jvm/src/main/scala, test/jvm/src/test/scala, test/shared/src/main/scala, test/shared/src/test/scala} from repository zio
Checking out jvm/src/main/scala from repository portable-scala-reflect
Problem with the build server: 'org.eclipse.lsp4j.jsonrpc.JsonRpcException: java.io.IOException: Pipe closed'.
  at java.io.PipedInputStream.checkStateForReceive(PipedInputStream.java:260)
  at java.io.PipedInputStream.receive(PipedInputStream.java:226)
[...]

PR #440, opened 2019-08-02
Fixed a bug in the pool that sometimes returned stale connections. 

odisseus:
Also removed the retrying pool because it's redundant now. 

This pull request resolves the issue #397.


propensive (approved):
So, does this fix the most annoying bug we've had to deal with?! 😊

odisseus:
So it seems. The connection might still break while being used (e. g. if the Bloop server suddenly crashes), but a known broken connection will never be reused.

PR #441, opened 2019-08-03
Change default install destination

propensive:
Install Fury into /opt/fury-$version if running as root, and into
$HOME/.fury if not, unless $FURYHOME is already set (i.e. it's a
reinstallation), in which case, use the old location.


etc/install.sh
```diff
@@ -68,61 +95,62 @@ prepare() {
...
-  cp "$DESTINATION/etc/fishrc" "$CONFIG/fish"
+  cp "${DESTINATION}/etc/fishrc" "${CONFIG}/fish"
 }
 
 updateShells() {
-  echo "Updating shell configuration"
-  mkdir -p "$CONFIG"
-  cp "$DESTINATION/etc/aliases" "$CONFIG/aliases"
+  message "Updating shell configuration"
+  mkdir -p "${CONFIG}"
```
odisseus, line 153: Looks like `$CONFIG` is not being initialized during the cloud build.

etc/install.sh
```diff
@@ -1,37 +1,53 @@
 #!/usr/bin/env bash
 FURY_VERSION=test
-DESTINATION="$HOME/fury-$FURY_VERSION"
-CONFIG="$HOME/.furyrc"
```
odisseus, line 4: I think this line should stay so that `$CONFIG` is guaranteed to have a sensible default value.

propensive:
@odisseus The reason this is slow is that running the tests in a different location causes the Bloop launcher to do a lot of downloading using Coursier which it has already done during the build. The `.cache` directory (containing Coursier's downloaded files) in `/home/bash_user/` starts off at 20MB, but is 104MB after the integration tests have run. The `.cache` directory in `/root/` is initially 55MB, so when the tests are run as root, it seems like there's less downloading to do. For some reason, that 35MB makes the difference.

PR #442, opened 2019-08-04
Reduce the amount of downloading required

propensive:


Dockerfile
```diff
@@ -1,44 +1,51 @@
...
 
 # Set up Git credentials
-RUN git config --global user.email 'fury@propensive.com' && git config --global user.name 'Fury Test'
+RUN (git config --global user.email 'fury@propensive.com' && git config --global user.name 'Fury Test')
 
 # Install Scala 2.12.8
-RUN mkdir /opt/scala-2.12.8 && \
-	curl -s https://downloads.lightbend.com/scala/2.12.8/scala-2.12.8.tgz | tar xz -C /opt/scala-2.12.8 --strip 1
+COPY pkg/* /build/
+RUN (if ! [ -f /build/scala.tar.gz ]; \
```
odisseus, line 10: Consider using `wget -nc`.

propensive: That would have been better, though I'm not sure if `wget` is as standard as `curl`. I have vague memories that I used it in a makefile before and it didn't work for someone on a Mac, though I'm not certain of that.

odisseus: At least you can add `--silent` to reduce spam in the log. Then again, why is there an uncertainty whether `/build/scala.tar.gz` is present at this point??? Isn't this command executed inside the image environment, which has a known state? 

_P. S._ Ah, I see it now — the makefile might or might not have put that archive in place. In that case, I think this check should be performed in the makefile itself so that the tarball is guaranteed to be present by the time the dockerfile is executed.

propensive:
I think I'll close this PR. Any objections, @odisseus?

odisseus:
Replacing `ADD` with `COPY` is a good idea and it would be good to merge those changes separately. As for the other changes in this pull request, I'm not sure.

_P.S._ The good changes have been merged in the PR #461, and the rest are questionable. I think this pull request should be closed without merging.

Issue #443, opened 2019-08-07
References to different commits in the same repository are considered equal


odisseus:
Consider the case class [`ExternalSource`](https://github.com/propensive/fury/blob/master/src/core/data.scala#L1451), which represents a reference to an external code repository. Two instances of this class will be considered equal if they reference the same remote source and are checked out into the same path. 

However, an `ExternalSource` can occur in multiple schemas in the same build, and every schema may resolve it to a different instance of [`SourceRepo`](https://github.com/propensive/fury/blob/master/src/core/data.scala#L1150), e. g. pointing to different commits on the same branch. 

These potential conflicts are not caught during the [hierarchy construction](https://github.com/propensive/fury/blob/master/src/core/data.scala#L840), because the merging algorithm compared only the repo references, not the commits they are resolved to.

odisseus:
https://github.com/propensive/fury/blob/6e2021577510548b77857431221e4a352a0e0955/src/core/data.scala#L780

This line assumes that every direct dependency of a target is contained in the same schema.

odisseus:
As of version [0.6.7](https://github.com/propensive/fury/releases/tag/v0.6.7), this problem seems to be fixed.

PR #444, opened 2019-08-08
Added a check whether different references to the same repository refer to the same commit.

odisseus:
This pull request fixes the issue #443.

**Warning!** This pull request exposes the inconsistencies in the current definition of the Fury build, which formerly went unnoticed. Until these inconsistencies are resolved, Fury will be unable to build itself.


odisseus:
I realized that this pull request introduces another kind of problem.

There might be legitimate reasons why different modules would need different commits from the same repository, and these commits won't necessarily be interchangeable in any way.

odisseus:
After the pull request #455 has been merged, this one shouldn't be needed anymore.

PR #446, opened 2019-08-09
Created an integration test for detection of version conflicts between references to the same repository

odisseus:
This pull request provides an integration test for detection of conflicts between references to the same repository, as described in the issue #443.


propensive (changes requested):
You need to do something at the end of the test which actually forces Fury to attempt to construct the Universe, which is where the conflict resolution should happen.

propensive (approved):

PR #447, opened 2019-08-10
[WIP] Allow addition of new permission requirements

propensive:
This is the start of support for a security layer for Fury, and adds support for requiring permissions to run certain application modules as part of their definition.

The commands, `fury permission require` and `fury permission obviate` can be used to add and remove the permissions necessary for each application module to run. This does not grant an application module those rights; it merely specifies what they are. A separate command, `fury permission grant` is provided to actually grant the permissions in the current environment (i.e. the current installation of Fury on a computer).

I'm experimenting with the best way to grant permissions, and the best ways to grant a permission to an application module, without needing to repeatedly re-grant it whenever something changes.

To do:
- [ ] `--all` flag on `fury permission grant` command to grant all permissions for a particular module
- [ ] Permission revocation
- [ ] Checking of permissions before launching compilation (currently they're checked only by the security manager)
- [ ] Option to grant a permission on the local machine at the same time as adding it to a module
- [ ] Provision of an optional "reason" field to explain why a permission is required

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ dist/fury-$(VERSION).tar.gz: all-jars dist/bundle/bin/fury dist/bundle/etc
 #TODO refactor etc structure (separate bundled scripts from development ones)
 dist/bundle/etc:
 	mkdir -p $@
-	cp -r etc/aliases etc/bashrc etc/fishrc etc/zshrc etc/completion etc/security $@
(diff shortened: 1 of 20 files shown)


propensive:
It's passing the tests now, so I'm going to leave the other actions for later PRs.

PR #451, opened 2019-08-12
Set up a shared Coursier cache for different users that use Coursier during the integration tests

odisseus:
Trying to share the Coursier [cache](https://github.com/propensive/fury/pull/441#issuecomment-520223770) between root and other users.


propensive:
@odisseus Did you notice if it's any faster? Is it ready to merge?

odisseus:
I haven't noticed any speedup, but reusing the cache should reduce network traffic. In the event of network issues, this will mean an actual speedup or maybe even success for a build that would otherwise time out.

propensive (approved):

PR #465, opened 2019-08-16
Allow --https to be specified as a cmdline parameter

propensive:


src/core/args.scala
```diff
@@ -59,6 +59,7 @@ object Args {
   val ForceArg = CliParam[Unit]('F', 'force, "force this operation")
   val FatJarArg = CliParam[Unit]('F', Symbol("fat-jar"), "package the module along with all its dependencies")
   val FileArg = CliParam[Path]('f', 'file, "destination file")
+  val HttpsArg = CliParam[Unit]('H', 'https, "use HTTPS to resolve repository aliases instead of SSH")
```
odisseus, line 62: I think this should be the default option.

Maybe the user shouldn't be able to choose between SSH and HTTPS at all for Github repositories; Fury would try both ways, starting with HTTPS.

propensive: It's my intention to make it part of the configuration, that is to say, the user could write:
```
fury config set --https
```
or similar, to make it the default everywhere.

Currently, in the places where `--https` can't be specified, such as when tab completions are executed, we're defaulting to SSH, though this ought to default to whatever is configured.

There's a question as to what should be the default for the configuration. I'm not sure.

odisseus: I think HTTPS should be the default way to resolve Github repositories, because it is [easier to set up](https://stackoverflow.com/questions/11041729/why-does-github-recommend-https-over-ssh). However, not all Git servers [support](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols) this way of connecting.

PR #489, opened 2019-08-22
Decoupled the definition of syncCompilation from asyncCompilation.

odisseus:


odisseus:
@propensive, I'm merging this because we have discussed these changes yesterday, and IIRC you have agreed to them.

PR #493, opened 2019-08-23
Added a test for concurrent borrowing of resources.

odisseus:


src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
       dummyPool.borrow("a/b/c"){void}
       dummyPool.borrow("a/b/x"){void}
       dummyPool.borrow("a/b/c"){void}
       dummyPool.size
     }.assert(_ == 2)
 
+    test("support concurrent requests") {
+      val dummyPool = new DummyPool
+      val allFinished = new CountDownLatch(20)
+      val keys = "klmnop".toSeq.map(c => s"a/b/$c")
```
mzarnowski: if the string `klmnop` stands for something meaningful, please extract it to a variable to avoid confusion.
If it does not - can you use something like "abcdef" or "uvwxyz"? Maybe the latter, since it doesn't look like anything on the first glance?

src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
-    override def create(key: String): Symbol = Symbol(key)
-    override def destroy(value: Symbol): Unit = ()
-    override def isBad(value: Symbol): Boolean = false
+
+  class Resource(val tag: String)
+
+  class DummyPool extends Pool[String, Resource](10L) {
+      override def create(key: String): Resource = new Resource(key)
+      override def destroy(value: Resource): Unit = ()
+      override def isBad(value: Resource): Boolean = false
```
mzarnowski, line 34: does `isBad` mean `isInvalid`? Or something entirely different?

src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
+      val allFinished = new CountDownLatch(20)
+      val keys = "klmnop".toSeq.map(c => s"a/b/$c")
+      (1 to 20).map { _ => Future {
+        Random.shuffle(keys).foreach{ key =>
+          dummyPool.borrow(key){void}
+          Thread.sleep(100)
+        }
+        allFinished.countDown()
+      }}
+      allFinished.await()
```
mzarnowski: is there a time limit somewhere, so that this test doesn't hang on failure?

src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
+      println(dummyPool.size)
```
mzarnowski:
```suggestion
```

src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
+      dummyPool.size
```
mzarnowski: it would be more expressive to have:
```scala
  (...)
  allFinished.await()
  dummyPool
}.assert(_.size == 6)
...

File: src/core-test/testpool.scala (hunk 1 above)
Comment (mzarnowski) [inline src/core-test/testpool.scala]: do those `/` matter? if not, maybe we could use `UUID.randomUUID`  to convey the lack of inherent meaning of the keys?

File: src/core-test/testpool.scala [hunk 6]
```diff
@@ -16,36 +16,59 @@
...
+      val dummyPool = new DummyPool
+      var result1: Resource = null
+      var result2: Resource = null
+      var result3: Resource = null
       dummyPool.borrow("a/b/c"){result1 = _}
       dummyPool.borrow("a/b/x"){result2 = _}
       dummyPool.borrow("a/b/c"){result3 = _}
-      (result1, result2, result3)
-    }.assert(_ == (Symbol("a/b/c"), Symbol("a/b/x"), Symbol("a/b/c")))
+      (result1.tag, result2.tag, result3.tag)
```
mzarnowski, line 67:
```suggestion
      (result1, result2, result3).map(_.tag)
```

mzarnowski (changes requested):
The test looks sound. There is a rouge `println` in there, so after removing it, the PR can be merged if this is urgent, as other comments are purely cosmetic.

src/core-test/testpool.scala
```diff
@@ -16,36 +16,59 @@
...
       dummyPool.size
     }.assert(_ == 2)
 
+    test("support concurrent requests") {
+      val dummyPool = new DummyPool
+      val allFinished = new CountDownLatch(20)
+      val keys = "klmnop".toSeq.map(c => s"a/b/$c")
+      (1 to 20).map { _ => Future {
+        Random.shuffle(keys).foreach{ key =>
+          dummyPool.borrow(key){void}
```
mzarnowski: this takes ownership and immediately returns it, right? 
Would the semantic hold, if the `Thread.sleep(100)` was moved in the place of `void` ?

src/core-test/testpool.scala
odisseus: Okay, if you really think that "uvwxyz" is more (or less?) meaningful than "klmnop"...

src/core-test/testpool.scala
odisseus: It just means that the resource cannot be used anymore, whatever the cause might be.

src/core-test/testpool.scala
odisseus: Yes, the unit tests would eventually [time out](https://github.com/propensive/fury/blob/master/cloudbuild.yaml#L14).

src/core-test/testpool.scala
odisseus: Ok

src/core-test/testpool.scala
odisseus: Does it really make a difference?

src/core-test/testpool.scala
odisseus: UUIDs are hard to compare visually. Besides, when the pool is used for storing BSP connections, the keys very much have meaning; they are paths to the corresponding working directories.

src/core-test/testpool.scala
odisseus: Would this even compile?

src/core-test/testpool.scala
odisseus: It should have been there all along.

src/core-test/testpool.scala
mzarnowski: semantically? no.

but with this change you don't have to look on the above line to know what is being compared with zero. I think the easier it is to reason about a piece of code, the better it is.

ideally it should be `pool.size == 0` but that is more boilerplatish with `probably`

src/core-test/testpool.scala
mzarnowski: OK, I agree UUID might be a bad choice here.

If I understand correctly, the pool itself doesn't care about the meaning of the keys, right? If so, the tests should also be as simple as possible. If the sequence `('a' to 'z').map(i => s"key $i")` works, then it should be used **unless** the test itself depend on the form of the key.

src/core-test/testpool.scala
mzarnowski: right, that is not a list but a tuple... nevermind then

src/core-test/testpool.scala
mzarnowski: so, basically, another word for `invalid` ;)

src/core-test/testpool.scala
mzarnowski: when replying to the comment below, I noticed that what really would be nicer is: `('k' to 'p')` ;)

odisseus: Ok

src/core-test/testpool.scala
odisseus: Ok

src/core-test/testpool.scala
odisseus: Ok

PR #501, opened 2019-08-30
Introduce cli completion

odisseus:
This is an attempt to rebase pull request #273.


mzarnowski (approved):
since the logic does not depend on `fury` itself, I believe it is OK as long as the tests are still passing :)

PR #504, opened 2019-09-03
Replaced event streams with iterators

odisseus:
Each compile event is processed by the graph printer only once. Therefore, there is no point to store past events in a stream; they may be discarded as soon as they are processed. This eliminates an opportunity for memory leaks and stack overflows.

### DIFF
--- a/src/core/graph.scala
+++ b/src/core/graph.scala
@@ -49,7 +49,7 @@ object Graph {
   def live(changed: Boolean,
            io: Io,
            graph: Map[ModuleRef, Set[ModuleRef]],
-           stream: Stream[CompileEvent],
+           stream: Iterator[CompileEvent],
            state: Map[ModuleRef, CompilationInfo],
            streaming: Boolean)
           (implicit theme: Theme)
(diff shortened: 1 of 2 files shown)


src/core/graph.scala
```diff
@@ -42,69 +42,81 @@ object Graph {
...
+  private case class GraphState(changed: Boolean,
+                        io: Io,
+                        graph: Map[ModuleRef, Set[ModuleRef]],
+                        stream: Iterator[CompileEvent],
+                        compilationLogs: Map[ModuleRef, CompilationInfo],
+                        streaming: Boolean){
+
+    def updateCompilationLog(ref: ModuleRef, f: CompilationInfo => CompilationInfo): GraphState = {
+      val previousState = compilationLogs.getOrElse(ref, CompilationInfo(state = Compiling(0), messages = List.empty))
+      val foo = compilationLogs.updated(ref, f(previousState))
```
mzarnowski: rename `foo`  Maybe `updatedState` ?

src/core/graph.scala
```diff
@@ -42,69 +42,81 @@ object Graph {
...
+
+  }
 
   @tailrec
+  private def live(graphState: GraphState)(implicit theme: Theme): Unit = {
+    import graphState._
+
+    io.print(Ansi.hideCursor())
+    if (stream.hasNext) {
+      stream.next match {
```
mzarnowski: `live` is repeated at the end of every branch, maybe it could be extracted?
```scala
val foo = stream.next match { 
  (...)
}
live(foo)

File: src/core/graph.scala [hunk 3]
```diff
@@ -42,69 +42,81 @@ object Graph {
...
+        case StartCompile(ref) =>
+          live(graphState.updateCompilationLog(ref, _.copy(state = Compiling(0))).copy(changed = true))
+        case DiagnosticMsg(ref, msg) =>
+          live(graphState.updateCompilationLog(ref, Lens[CompilationInfo](_.messages).modify(_)(_ :+ msg)).copy(changed = false))
+        case NoCompile(ref) =>
+          val newState = compilationLogs.getOrElse(ref, CompilationInfo(state = AlreadyCompiled, messages = List.empty))
+          live(graphState.updateCompilationLog(ref, _ => newState).copy(changed = true))
+        case StopCompile(ref, success) =>
+          val msgs = compilationLogs(ref).messages
+          val newState = CompilationInfo(if (success) Successful(None) else Failed(""), msgs)
```
mzarnowski, line 85: what does the `Failed("")` mean and where (if anywhere) would the user see an empty string?

mzarnowski (changes requested):
Using iterator sounds like a good idea.
There are minor things which can be improved before merging.

src/core/graph.scala
odisseus: I'll just inline it.

src/core/graph.scala
odisseus: Ok

src/core/graph.scala
odisseus: This is a token that makes the graph printer [print](https://github.com/propensive/fury/blob/master/src/core/graph.scala#L154) "_x_ errors". Its field is just ignored.

mzarnowski (approved):

Issue #512, opened 2019-09-13
Simplified interface for granting permissions


odisseus:
The command that defines a new permission should print out its identifier (hash), so that the user can grant or obviate that permission with the next command.

The command that defines a new permission should also grant it immediately. There should be a flag to define a new permission without granting it.

PR #513, opened 2019-09-13
Simplified interface for granting permissions

odisseus:
This pull request resolves the issue #512.


propensive (commented):
Is it currently possible to add a permission using its full-length hash (i.e. the one that gets printed) rather than the short version of the hash?

odisseus:
Nope

propensive:
Ok, we can add this in a separate PR. This one looks good, thank you! Once CI goes green, of course.

odisseus:
I think that's a serious issue, and I'm going to try to fix it.

PR #515, opened 2019-09-13
Don't use `-D` parameter with `install`.

propensive:


etc/fury
```diff
@@ -91,8 +91,8 @@ ngServer() {
     SCALA=$(coursier fetch --classpath org.scala-lang:scala-reflect:2.12.8 com.facebook:nailgun-server:1.0.0)
     java -Dfury.home="${FURYHOME}" -cp "$SCALA:$CLASSPATH" "$MAIN" "$PORT" &
     NG_PID=$!
-    install -D <(echo $NG_PID) $PIDFILE
```
odisseus, line 94: How do you make sure that all the parent directories of `$PIDFILE` exist in the file system?

propensive: I ended up removing the use of a PID file completely.

etc/fury
```diff
@@ -82,22 +82,26 @@ stopFury() {
   nailgun "${FURY_MAIN}" "stop" && killFury
 }
 
+alive() {
+  bash -c "(exec 6<> /dev/tcp/127.0.0.1/${PORT} > /dev/null 2> /dev/null)"
```
odisseus, line 86: I'm afraid this won't work in some situations. For example, what about the moment when the Nailgun process is already running, but hasn't started listening for connections yet? If the launcher script is run multiple times in quick succession (as happens during the integration tests), this situation is almost certain to happen.

odisseus: Nevermind, this turned out to work as intended.

Issue #521, opened 2019-09-16
Cannot use a compiler module defined in an external schema


odisseus:
This was probably introduced by pull request #485.

## Steps to reproduce
```
fury layer init
fury repo add -u gh:propensive/.scala -n scala
fury import add -i scala:scala-2.12.8
fury project add -n hello-world
fury module add -n app -c scala/compiler -t application -M HelloWorld
echo $?
```

PR #522, opened 2019-09-16
Fixed resolution of compiler references.

odisseus:
This pull request fixes the issue #521.


propensive (approved):
Looks like this ended up being more complex than we thought...

PR #551, opened 2019-09-28
Support Bloop pipelining

propensive:


propensive:
Pipelining causes multi-module builds to produce output in a different order. It's probably still stable, though.

odisseus:
Looks good, but the integration tests that have multiple modules are now broken.

src/core/args.scala
```diff
@@ -74,6 +74,7 @@ object Args {
   val LicenseArg = CliParam[LicenseId]('L', 'license, "license for code in this project")
   val ModuleArg = CliParam[ModuleId]('m', 'module, "specify a module")
   val MainArg = CliParam[String]('M', 'main, "specify a main class")
+  val PipeliningArg = CliParam[Boolean]('p', Symbol("pipelining"), "use compilation pipelining (on, off) (experimental)")
```
odisseus: This argument has the same letter as `ProjectArg`, and some commands accept both of them.

odisseus:
While rebuilding this pull request with itself, I got the following error:
```
[E] fury/schema>local:src/schema/schema.scala +91:
  | not found: value 
  | rows      <- ~Diff.gen[Schema].diff(schema, other)
```
When it happened, the compilation of `fury/core` was not completed yet.

Rebuilding the same code with make gave no errors.

propensive:
I think that the failure is acceptable. It's probably a bug in Bloop, and it fits with the decision to mark pipelining as "experimental".

PR #554, opened 2019-09-29
Store layers centrally

propensive:
This stores layers in `~/.furyrc/layers/<hash>` with a `.focus.fury` file which points to the current layer. It's currently work-in-progress.


odisseus:
Does this pull request provide an easy way to compare the current state of the layer with the state from the latest commit? At the moment this can be achieved by simple `git diff`, but the diff will become useless if the layer file is replaced with a hash.

By the way, will it be possible to check out a complete project from a Git repository?

src/model/ids.scala
```diff
@@ -427,8 +427,8 @@ object SchemaId {
   final val default = SchemaId("default")
 
   def parse(name: String): Try[SchemaId] = name match {
-    case r"[a-z](-?[a-zA-Z0-9]+)*" => Success(SchemaId(name))
-    case _                         => Failure(InvalidValue(name))
+    case r"[a-z]([-\.]?[a-zA-Z0-9]+)*" => Success(SchemaId(name))
```
odisseus, line 430: Is this a fix that applies to the current data model as well?

propensive:
I think this got merged via another pull request.

PR #557, opened 2019-09-30
Implemented continuous rebuilding

odisseus:
Linear output works well; in normal mode output and the graph break each other.

### DIFF
--- a/layer.fury
+++ b/layer.fury
@@ -16,7 +16,10 @@ schemas	default	id	default
 								hidden	false
 						params	
 						sources	src/build	src/build
-						binaries	
+						binaries	io.methvin:directory-watcher-better-files_2.12:0.9.6	binRepo	central
+								group	io.methvin
+								artifact	directory-watcher-better-files_2.12
+								version	0.9.6
 						resources	
 						bloopSpec	None
 						environment	
--- a/src/build/rebuild.scala
+++ b/src/build/rebuild.scala
@@ -1,0 +1,110 @@
+/*
+   ╔═══════════════════════════════════════════════════════════════════════════════════════════════════════════╗
+   ║ Fury, version 0.6.7. Copyright 2018-19 Jon Pretty, Propensive OÜ.                                         ║
+   ║                                                                                                           ║
+   ║ The primary distribution site is: https://propensive.com/                                                 ║
+   ║                                                                                                           ║
+   ║ Licensed under  the Apache License,  Version 2.0 (the  "License"); you  may not use  this file  except in ║
+   ║ compliance with the License. You may obtain a copy of the License at                                      ║
+   ║                                                                                                           ║
+   ║     http://www.apache.org/licenses/LICENSE-2.0                                                            ║
+   ║                                                                                                           ║
+   ║ Unless required  by applicable law  or agreed to in  writing, software  distributed under the  License is ║
+   ║ distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. ║
+   ║ See the License for the specific language governing permissions and limitations under the License.        ║
+   ╚═══════════════════════════════════════════════════════════════════════════════════════════════════════════╝
+*/
+package fury
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import _root_.io.methvin.better.files.RecursiveFileMonitor
+import better.files.File
+import fury.core._
+import fury.io.Path
+import fury.model._
+import fury.strings._
+import fury.utils.Multiplexer
+
+import scala.concurrent.{Await, ExecutionContext, Future, duration}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object RebuildService {
+  def repeatBuild(io: Io, compilation: Compilation, project: Project, module: Module, layout: Layout, 
+                  policy: Policy, reporter: Reporter, theme: Theme, compileArgs: List[String]): Unit = {
+    val multiplexer = new Multiplexer[ModuleRef, CompileEvent](compilation.targets.map(_._1).to[List])
+    val allDirectories = compilation.targets.values.flatMap(_.sourcePaths).toSet
+    io.println(str"Watching ${allDirectories.size} directories...")
+    val sourceWatcher = new SourceWatcher(io, allDirectories)
+    val future = repeat(
+      io,
+      skip = () => !sourceWatcher.hasChanges,
+      action = () => buildOnce(io, compilation, project, module, layout, policy, compileArgs, sourceWatcher, multiplexer)
+    )
+    reporter.report(io, compilation, theme, multiplexer)
+    val compileResults = Await.result(future, duration.Duration.Inf)
+    multiplexer.closeAll()
+    sourceWatcher.stop()
+  }
+  
+  private def buildOnce(io: Io, compilation: Compilation, project: Project, module: Module, 
+                         layout: Layout, policy: Policy, compileArgs: List[String],
+                        sourceWatcher: SourceWatcher, multiplexer: Multiplexer[ModuleRef, CompileEvent]) = {
+    sourceWatcher.clear()
+    val moduleRef = module.ref(project)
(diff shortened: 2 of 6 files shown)


layer.fury
```diff
@@ -16,7 +16,14 @@ schemas	default	id	default
 								hidden	false
 						params	
 						sources	src/build	src/build
-						binaries	
+						binaries	com.github.pathikrit:better-files_2.12:3.8.0	binRepo	central
+								group	com.github.pathikrit
+								artifact	better-files_2.12
+								version	3.8.0
```
propensive: Do we actually need to include this? Isn't it a transitive dependency of `directory-watcher-better-files`?

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+private class SourceWatcher(io: Io, sources: Set[Path]) {
+  private val changed = new AtomicBoolean(true)
+
+  private val watchers = sources.map( src => new RecursiveFileMonitor(File(src.value)) {
+    override def onCreate(file: File, count: Int) = onChange(file)
+    override def onModify(file: File, count: Int) = onChange(file)
+    override def onDelete(file: File, count: Int) = onChange(file)
+
+    override def start()(implicit executionContext: ExecutionContext): Unit = {
+      this.watcher.watchAsync(SourceWatcher.executor)
```
propensive: Why do you refer to `this` explicitly?

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+  def hasChanges: Boolean = changed.get
+
+  def clear: Unit = changed.set(false)
+
+  def stop: Unit = watchers.foreach(_.stop)
+
+}
+
+object SourceWatcher{
+  val executor = java.util.concurrent.Executors.newCachedThreadPool()
```
propensive, line 107: I'm starting to think that we have too many thread pools, fork/join pools and `ExecutionContext`s littered around the code. I don't think anything needs to change for this PR, but can you open an issue to clean these up?

src/frontend/menu.scala
```diff
@@ -41,7 +41,8 @@ object FuryMenu {
             Action('compile, msg"compile a module", BuildCli.compile(None, None)),
             Action('describe, msg"describe the build for a module", BuildCli.describe),
             Action('save, msg"save a JAR file", BuildCli.save),
-            Action('native, msg"build a native executable using GraalVM", BuildCli.native)
+            Action('native, msg"build a native executable using GraalVM", BuildCli.native),
+            Action('watch, msg"watch the sources and rebuild on every change", BuildCli.watch)
```
propensive, line 45: Would it be possible to make this a `--watch` flag on the `compile` command, instead?

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+    io.println(str"Recompiling $moduleRef")
+    val compilations = compilation.compile(io, moduleRef, multiplexer, Map(), layout, policy, compileArgs)
+    Future.sequence(compilations.values)
+  }
+
+  private def repeat[Res](io: Io, skip: () => Boolean, action: () => Future[Res]): Future[Res] = {
+    for{
+      x <- if (skip()) {
+        io.print(Ansi.up(1)())
+        io.println(str"No changes, waiting...")
```
propensive: Could this just say "Waiting for changes..."?

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+    val moduleRef = module.ref(project)
+    io.println(str"Recompiling $moduleRef")
+    val compilations = compilation.compile(io, moduleRef, multiplexer, Map(), layout, policy, compileArgs)
+    Future.sequence(compilations.values)
+  }
+
+  private def repeat[Res](io: Io, skip: () => Boolean, action: () => Future[Res]): Future[Res] = {
+    for{
+      x <- if (skip()) {
+        io.print(Ansi.up(1)())
```
propensive, line 64: For the graph reporter, at least, I think we would like to clear the entire screen when a new build is triggered. `\033[2J` should do it.

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+        io.println(str"No changes, waiting...")
+        Thread.sleep(1000)
```
propensive, line 66: Can you include a comment to explain this?

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+        repeat(io, skip, action)
```
propensive, line 67: Is this method tail-recursive? If not, too many repetitions and it will blow the stack.

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+    multiplexer.closeAll()
+    sourceWatcher.stop
+  }
+  
+  private def buildOnce(io: Io, compilation: Compilation, project: Project, module: Module, 
+                         layout: Layout, policy: Policy, compileArgs: List[String],
+                        sourceWatcher: SourceWatcher, multiplexer: Multiplexer[ModuleRef, CompileEvent]) = {
+    sourceWatcher.clear
+    val moduleRef = module.ref(project)
+    io.println(str"Recompiling $moduleRef")
```
propensive: This should be `msg""` instead of `str""`

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+        action()
+      }
+      _ = io.println("Rebuild complete")
+      y <- if(Thread.currentThread.isInterrupted) Future.successful(x) else repeat(io, skip, action)
+    } yield y
+  }
+}
+
+private class SourceWatcher(io: Io, sources: Set[Path]) {
+  private val changed = new AtomicBoolean(true)
```
propensive: `private[this]`

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+      io.println(str"File modified: ${Path(file.path)}")
+      changed.set(true)
+    }
+  }
+
+  watchers.foreach(_.start()(SourceWatcher.ec))
+
+  def hasChanges: Boolean = changed.get
+
+  def clear: Unit = changed.set(false)
```
propensive: Should be `clear()` instead of just `clear` to indicate that it is side-effecting.

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  def stop: Unit = watchers.foreach(_.stop)
```
propensive: Same for this: `stop()`.

src/build/rebuild.scala
```diff
@@ -0,0 +1,111 @@
...
+      x <- if (skip()) {
+        io.print(Ansi.up(1)())
+        io.println(str"No changes, waiting...")
+        Thread.sleep(1000)
+        repeat(io, skip, action)
+      } else {
+        action()
+      }
+      _ = io.println("Rebuild complete")
+      y <- if(Thread.currentThread.isInterrupted) Future.successful(x) else repeat(io, skip, action)
```
propensive, line 72: Can you let me know how thread interruption works some time? This is probably easier in-person!

propensive (changes requested):

src/build/rebuild.scala
odisseus: Ok

src/build/rebuild.scala
odisseus: Ok

src/build/rebuild.scala
odisseus: Ok

src/build/rebuild.scala
odisseus: Ok

src/build/rebuild.scala
odisseus: Ok

layer.fury
odisseus: Removed

src/build/rebuild.scala
odisseus: Ok

odisseus:
See #561.

Issue #539, opened 2019-09-25
Watch the local sources for changes and rebuild them as needed


odisseus:
Here are the actions for which continuous re-running definitely makes sense:
* compilation (`fury build compile`)
* execution of application modules
* compilation and packaging (`fury build save`)
* compilation to binary (`fury build native`)

I'm not so sure about the commands that modify the layer (such as adding or removing dependencies). On the one hand, those commands definitely can affect the build output; on the other hand, they can alter the build drastically — for example, remove the very module that has been put on continuous rebuild!

odisseus:
https://github.com/jvican/directory-watcher
https://gitter.im/scalacenter/bloop?at=5d92137424509b6046857cb2

PR #559, opened 2019-10-02
Unified the code responsible for compilation tasks.

odisseus:
This refactoring is required for a better implementation of continuous rebuild, and thus is related to the issue #539.


propensive:
Deletions outnumber additions. That's a good sign. :)

PR #576, opened 2019-10-10
Remove launcher

odisseus:


odisseus:
Bloop 1.3.4 employs [classpath tricks](https://github.com/scalacenter/bloop/blob/master/build.sbt#L300). This means that we probably have to build our own version of the launcher.

PR #586, opened 2019-10-11
Daemonize all the things

odisseus:
This makes the stop command functional.
After a continuous rebuild had been launched, the stop command doesn't work, but the kill command works even without the force switch.

Both commands leave the Bloop server in place, terminating just the Nailgun process.


propensive (approved):

propensive:
This looks like a neat solution!

propensive:
I think that `kill` (without `--force`) and `stop` ought to do the same thing. I see that one of them in the run script is using `ng-stop` and one is using just `stop`. I think one of these is incorrect. Could you also change that, @odisseus?

PR #613, opened 2019-10-15
Ipfs merged

odisseus:


propensive (approved):
If the compile errors are fixed, that's fantastic. I'll try to get it to a mergeable state tomorrow. That will primarily involve providing a "layer export" facility, which (I think) gets us back to more or less the same functionality we had before, regardless of what IPFS is doing.

propensive:
The unit tests pass, but CI still times out. I'm going to merge the PR and try to fix that later.

PR #627, opened 2019-10-16
Improved output of the build-zio test.

odisseus:
The output of the compile phase will be useful to locate the causes of possible compile errors. The output lines had to be sorted, because the project contains modules that don't depend on each other and, therefore, may be built in any order.

### DIFF
--- a/test/passing/build-zio/check
+++ b/test/passing/build-zio/check
@@ -1,1 +1,34 @@
+Checking out * from repository scala
+Checking out core/jvm/src/main/java from repository zio
+Checking out jvm/src/main/java from repository portable-scala-reflect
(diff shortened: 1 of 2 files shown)


propensive:
Yes, sorting the output is going to be a useful trick in more of these integration tests.

odisseus:
Actually, I want to avoid modifying the output wherever possible. Changing the order of the output can yield especially confusing results when something fails — imagine a sorted stack trace!

However, there is a conceptual difference between the rest of our integration tests and this one. In this test, we don't build the layer from scratch, neither do we actually care about precise output. The important result here is the correct artifacts built in reasonable time.

I hope we will have more tests like this one in the future (maybe starting with Cats). But then we would need a different mechanism to run and check those tests.

PR #632, opened 2019-10-21
Append to PATH instead of prepending

kubukoz:
Reasoning: fury-installed bloop/coursier/whatnot should probably not cover up the programs already available on the PATH.


odisseus:
This change is reasonable, because the user might have Coursier and Nailgun already installed, and Fury would obscure them with its own versions. 

@propensive, is there any reason why the versions provided with Fury should take priority?

PR #633, opened 2019-10-22
Add module param to binary add command

### LINKED ISSUE #260: binaries can only be added to the current module
Commands like `fury binary add` do not have a `--module`/`-m` parameter, so you have to change the current module in order to add a binary to it. This isn't ideal.

IvoSaraiva:
The -m option was being supplied in a cli that was not being used neither in the BinariesCtx nor the ParamCtx. Moved it to the ModuleCli.context. 

This should fix #260 for binaries and also for params.

### DIFF
--- a/src/module/module.scala
+++ b/src/module/module.scala
@@ -51,7 +51,7 @@ object ModuleCli {
     cli          <- cli.hint(ProjectArg, schema.map(_.projects).getOrElse(Nil))
     optProjectId <- ~schema.flatMap { s => cli.peek(ProjectArg).orElse(s.main) }
     optProject   <- ~schema.flatMap { s => optProjectId.flatMap(s.projects.findBy(_).toOption) }
-    cli         <- cli.hint(ModuleArg, optProject.to[List].flatMap(_.modules))
+    cli          <- cli.hint(ModuleArg, optProject.to[List].flatMap(_.modules))
   } yield Context(cli, layout, config, layer, schema, optProject)
 
   def select(ctx: Context): Try[ExitStatus] = {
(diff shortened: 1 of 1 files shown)


src/module/module.scala
```diff
@@ -51,6 +51,7 @@ object ModuleCli {
     cli          <- cli.hint(ProjectArg, schema.map(_.projects).getOrElse(Nil))
     optProjectId <- ~schema.flatMap { s => cli.peek(ProjectArg).orElse(s.main) }
     optProject   <- ~schema.flatMap { s => optProjectId.flatMap(s.projects.findBy(_).toOption) }
+    cli          <- cli.hint(ModuleArg, optProject.to[List].flatMap(_.modules))
```
odisseus: If this is the right place to generate a hint for the `--module` argument, the value associated with this argument should also be retrieved here, like it is done with `--schema` and `--project`. Consider adding `optModule` as another field to the case class `ModuleCli.Context`.

By the way, doesn't this change enable the hint for `--module` argument even for commands `module add` and `module list`? These two commands don't (and shouldn't) accept that argument.

odisseus (changes requested):

src/module/module.scala
IvoSaraiva: You're right on the module commands problem. I'll try again.

IvoSaraiva:
What about just updating the MoculeCli.Context.cli in the BinariesCtx and ParamCtx? All binaries and params commands should accept a --module option.

propensive (approved):

odisseus:
I'm OK with merging this if @propensive approves. But the commits need to be squashed, because there are 5 commits that change just two lines in total.

propensive:
Yes, I just tried it out, and it works great. Nice that the fix was so simple!

PR #641, opened 2019-11-05
Cache Coursier-downloaded files in separate directory

propensive:
This will use Coursier to download binaries, as usual, but instead of using
an in-memory cache for these files, they are now stored in a directory with
the name,
```
<workspace>/.fury/binaries/<group>/<artifact>/<version>
```

If the directory exists, it is considered to have been a successful 
download, and to contain a correct set of binary JARs.

Fury currently keeps these files inside each project workspace, but they 
could (and probably should) be stored globally.

Fury's store for these files uses hard links to the Coursier directory to
reduce duplication of data.

We should check what happens if we attempt to create hard links across 
different filesystems, as might happen if Coursier's cache isn't on the 
same disk as Fury's. We should fall back to a copy operation, if necessary.

The motivation for this is that Coursier will currently fail if the 
computer is offline, because it does a re-check over the network. This 
should also make it faster using Fury for the first time, and should remove
one difference between running it standalone and as a daemon.


odisseus:
I don't like the idea of having another artifact cache. I would rather make use of the cache maintained by Coursier.

If Coursier fails when there is no network but the file is already cached, this is an issue of Coursier and it should be fixed by a pull request to Coursier.

propensive:
There's a small motivation to have Fury store its own copy of the JAR files it uses: it means that Fury is in control of everything it needs to perform a build. If Fury's cache directory is on the same physical device as Coursier's, then it doesn't use any extra space; it uses hard links. But if they're on different devices, there might be a significant waste of disk space.

PR #647, opened 2019-11-06
Catch JsonRpcException when closing a BSP connection

odisseus:


propensive (commented):
Looks like a good move.

PR #649, opened 2019-11-06
Experimentation around improving BSP connections

propensive:


odisseus (commented):
Is this pull request still relevant?

PR #650, opened 2019-11-06
wip

odisseus:


propensive (commented):
Looks like an improvement. Does it make any difference?

odisseus:
Nope

odisseus:
Now it should work

PR #657, opened 2019-11-12
Restored the sparse checkout function to the way it looked before the IPFS merge.

odisseus:
This works with Git 2.21.0.


propensive (approved):
Great. I hope this will clear up a few error messages I've seen, intermittently.

PR #682, opened 2019-11-13
Globalize installation

propensive:


src/model/layout.scala
```diff
@@ -42,50 +45,83 @@ object Layout {
   }
 }
 
-case class Installation(env: Environment) {
+object Xdg {
+  
+  private case class Var(name: String) {
+    private val variable = Option(System.getenv(s"XDG_$name"))
```
odisseus, line 51: Does this pull request include the necessary permissions to access these environment variables?

propensive: This code isn't run *by* Fury (which checks those permissions)... it's *part of* Fury, so we don't need to worry about permissions...

Or have I misunderstood the point?

odisseus: This code is invoked when `test-core` is executed. This is probably the reason why unit tests always failed for this pull request.

propensive: Ah, perfect! Well-spotted!

Makefile
```diff
@@ -106,9 +107,10 @@ fury-native: dist/bundle/lib/fury-frontend.jar
 	native-image -cp $(shell bash -c "ls $(NATIVEJARS) | paste -s -d: -") fury.Main
 
 test:
-	fury build run --https --output linear --project fury --module test-strings
-	fury build run --https --output linear --project fury --module test-ogdl
-	fury build run --https --output linear --project fury --module test-core
+	mkdir tests
+	fury layer extract -f build.fury -d tests
```
odisseus, line 122: I'm afraid this won't work without corresponding changes to the layer definition. 

In the current master, the layer defines source directories as relative paths. As a result, the build launched in the `tests`directory will look for sources in the subdirectories of `tests/src`, which just don't exist.

Makefile
```diff
@@ -106,9 +107,10 @@ fury-native: dist/bundle/lib/fury-frontend.jar
...
+	fury permission grant --module test-core --project fury -P 4fb
```
odisseus, line 123: Does this permission apply to the layer located in the current directory or to the one in `tests`?

PR #683, opened 2019-11-14
Support for relative navigation between layers

propensive:


src/model/ids.scala
```diff
@@ -73,22 +73,36 @@ case class ModuleId(key: String) extends Key(msg"module")
...
   def prefix(importId: ImportId): ImportPath =
-    ImportPath((importId :: parts).map(_.key).mkString("./", "/", ""))
+    ImportPath((importId :: parts).map(_.key).mkString("/", "/", ""))
+
+  def dereference(relPath: ImportPath): ImportPath =
+    if(relPath.path.startsWith("/")) relPath
+    else {
+      relPath.path.split("/").to[List] match {
+        case ".." :: tail => ImportPath(path.split("/").to[List].dropRight(1).mkString("/")).dereference(ImportPath(tail.mkString("/")))
+        case xs => ImportPath((path.split("/").to[List] ++ xs).mkString("/"))
```
odisseus: The value `path.split("/").to[List]` can be extracted into a variable.

src/model/ids.scala
```diff
@@ -73,22 +73,36 @@ case class ModuleId(key: String) extends Key(msg"module")
...
+  val Root: ImportPath = ImportPath("")
 
   // FIXME: Actually parse it and check that it's valid
   def parse(str: String): Option[ImportPath] =
     Some(ImportPath(str))
 }
 
 case class ImportPath(path: String) {
-  def parts: List[ImportId] = path.split("/").to[List].tail.map(ImportId(_))
+  def parts: List[ImportId] = path.split("/").to[List] match {
```
odisseus: If `path` doesn't contain slashes, this method will return `Nil`. Is this the intended behaviour?

propensive: Yes, I think that's correct. I had to special-case the `Nil` case, because the `split` method has inconsistent behavior for the empty string.

src/model/ids.scala
odisseus: What about lines 105 and 106?

propensive: What in particular about them?

odisseus: They still contain the same expression, whereas `rawParts` could be used instead.

odisseus (approved):
I think this pull request can be merged, but it would be nice to have an integration test for the new functionality.

src/model/ids.scala
propensive: I was confused: these lines only contain closing braces.

PR #689, opened 2019-11-15
Unarchived bootstrapping layers

odisseus:
The binary storage format for the layer that are needed to bootstrap Fury causes a lot of problems.

1. When switching between branches, it is easy to forget to refresh the focus file. As a result, the code will be built with a wrong build definition (wrong dependencies, binary versions etc.)
2. It is not clear whether the currently focused layer is the same one as the layer stored inside the binary file.
3. There is currently no way to compare different versions of the bootstrapping layer.
4. It is impossible to compare versions of the layer between two branches without checking out both of them at the same time.

This pull request aims to solve all of them by storing the bootstrapping set of layers in an unarchived form. The makefile packages them and passes the fresh archive to `fury extract`.

This pull request has a problem of its own. Since the name of the layer file depends on its contents, even a small change to the build definition can result in a huge diff. But at least the text files can be compared manually without special tools.


propensive:
Seems like a sensible change. I think we can do better with merging in future, but this will be useful for now.

PR #690, opened 2019-11-15
Globalize installation 2

odisseus:


etc/testshell.sh
```diff
@@ -31,7 +31,7 @@ esac
 
 apt-get -qq install "${SHELL_NAME}" > /dev/null
 useradd -s "$(which ${SHELL_NAME})" -G coursier -d "$NEW_HOME" "$NEW_USER"
-mkdir -p "$NEW_HOME"
+mkdir -p "$NEW_HOME" "$NEW_HOME/.cache" "$NEW_HOME/.local/share"
```
propensive, line 34: This shouldn't be necessary, should it?

etc/integration
```diff
@@ -42,6 +42,9 @@ runtest() {
   return $SUCCESS
 }
 
+mkdir -pv ~/.local/share/fury/layers
+mkdir -pv ~/.cache/fury/policies
+
```
propensive, line 47: These should be created by Fury. If they're not, then there's a problem elsewhere we need to fix...

propensive (changes requested):

etc/integration
odisseus: That's the problem I mentioned. If these directories don't exist, the build [fails](https://console.cloud.google.com/cloud-build/builds/40c9982d-15f3-46ff-aa7b-1531a51bf87f?project=integration-223912&utm_campain=github_app&utm_medium=link&utm_source=github#step_2).

src/model/layout.scala
```diff
@@ -42,50 +44,80 @@ object Layout {
...
+  
+  private case class Var(name: String) {
+    private val variable = Option(System.getenv(s"XDG_$name"))
+    def path: Option[Path] = variable.map(Path(_))
+    def paths: Option[List[Path]] = variable.map(_.split(":").to[List].map(Path(_)))
+  }
+  
+  val home: Path = Option(System.getenv("HOME")).map(Path(_)).getOrElse(Path("/"))
+  val cacheHome: Path = Var("CACHE_HOME").path.getOrElse(home / ".cache")
+  val dataHome: Path = Var("DATA_HOME").path.getOrElse(home / ".local" / "share")
```
propensive, line 56: I think we should add `.extant()` to the end of each of these lines, and remove the explicit creation of these directories, above.

PR #691, opened 2019-11-18
Use Scala 2.12.10 instead of Scala 2.12.8

propensive:


odisseus:
Looks like you forgot to update the layer file. Here is an easy way to do it:
```sh
fury layer export -f foo.tar.gz
tar -xzvf foo.tar.gz
rm foo.tar.gz
git diff .focus.fury # find out the hash of the old layer
rm layers/<old layer hash>
git add layers/
git diff --staged #the diff should be small
```

propensive:
I'm going to recreate this PR.

PR #693, opened 2019-11-18
Configured build step order for the cloud build.

odisseus:
https://cloud.google.com/cloud-build/docs/configuring-builds/configure-build-step-order


propensive:
Fantastic! Maybe we can find some more parallelism...

PR #695, opened 2019-11-19
Display "linear" output only when running inside Docker/CI

propensive:


propensive:
This PR does what it's meant to, but the CI tests are currently broken.

PR #697, opened 2019-11-19
Much-reduced Dockerfile

propensive:


propensive:
There wasn't anything useful in here any more.

PR #698, opened 2019-11-19
Small improvements

odisseus:
This pull request corrects the prerequisites for the test goal, as well as makes application modules fail fast if the policy file cannot be created.

### DIFF
--- a/src/core/shell.scala
+++ b/src/core/shell.scala
@@ -48,7 +48,7 @@ case class Shell(environment: Environment) {
       Environment((environment.variables ++ env).updated("SHARED", layout.sharedDir.value), environment.workDir)
 
     val policyFile = layout.policyDir / UUID.randomUUID().toString
-    policy.save(policyFile)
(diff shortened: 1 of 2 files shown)


src/core/shell.scala
```diff
@@ -48,7 +48,7 @@ case class Shell(environment: Environment) {
       Environment((environment.variables ++ env).updated("SHARED", layout.sharedDir.value), environment.workDir)
 
     val policyFile = layout.policyDir / UUID.randomUUID().toString
-    policy.save(policyFile)
+    policy.save(policyFile).get
```
propensive, line 51: Can't we `flatMap` over the result of `policy.save` somewhere? It's a `Try`, right?

odisseus: It is definitely possible, but the change wouldn't be small anymore. 

Besides, the next three methods up the call hierarchy use either `flatMap` over `Future` or none at all.

PR #699, opened 2019-11-20
Some changes which appear to fix BSP connections breaking

propensive:


propensive:
Looks like I accidentally checked in a reference to a file on my hard disk...

PR #700, opened 2019-11-20
Experiments with launcher synchronization

odisseus:


propensive:
Can we get a better PR name, please? ;)

PR #701, opened 2019-11-20
Added test for Scala 3 compilation

propensive:


propensive:
This includes a very basic test which proves that compilation is possible with Scala 3. (Currently using version 0.20.)

Many thanks to @tgodzik for help resolving the build file problems!

test/passing/scala3/src/hw.scala
```diff
@@ -0,0 +1,3 @@
+object HelloWorld extends App {
```
odisseus: The test would be better if the source file could not be compiled with older versions of Scala. This would make it clear that the new version is actually used.

propensive: Yes, agreed.

odisseus (approved):

odisseus:
This can be merged as soon as the integration tests run at least once without timing out.

PR #705, opened 2019-11-24
Use `boot` instead of `bootstrap`

propensive:


Makefile
```diff
@@ -58,17 +58,17 @@ dist/bundle/etc:
 
 # Compilation
 
-bootstrap/scala:
+boot/scala:
```
odisseus, line 61: I don't think this is a good change. [Bootstrapping](https://stackoverflow.com/a/1254561) is a known term, and it describes exactly what these tasks do: build a new version of the program using an older version of the same program.

PR #708, opened 2019-11-25
Use a more complete JSON library that uses Magnolia for generic derivations

propensive:
This will be particularly useful for doing other things with JSON, which we are likely to need soon.


.focus.fury
```diff
@@ -1,2 +1,2 @@
-layerRef	85D48E3D72DC775FE785EA3B2520DAD39A79C3D76D2540C864D47329BE02897B
-path	.
+layerRef	B9328F0D8074B861712A20F824FCDCC5EE101673922A34D884E734CEA508F3E7
+path	/
```
odisseus, line 2: What does this change mean?

Makefile
```diff
@@ -1,6 +1,6 @@
 VERSION=${shell sh -c 'cat .version 2> /dev/null || git --git-dir git/fury/.git describe --exact-match --tags 2> /dev/null || git --git-dir git/fury/.git rev-parse --short HEAD'}
 BLOOPVERSION=1.3.5
-FURYSTABLE=0.7.12
+FURYSTABLE=0.7.11
```
odisseus: This should be updated

layers/170FFB493AD839E66CD7F53CC22E808852B98777274C1F200EACE8BBA19AD952
```diff
@@ -1072,6 +1145,12 @@ schemas	default	id	default
 		imports	0A24A1586682A39838622B66BACF81A3B51A7F3448371AB31B54B75DA12DCD28:scala-2.12.8	id	scala
 				layerRef	0A24A1586682A39838622B66BACF81A3B51A7F3448371AB31B54B75DA12DCD28
 				schema	scala-2.12.8
-		main	Some	probably
+				follow	None
+			72BC2FA87BC7E554FABB094FF86E87B8C4FC8CC43DB9543C9986F46F82AB7E04:default	id	jawn
+				layerRef	72BC2FA87BC7E554FABB094FF86E87B8C4FC8CC43DB9543C9986F46F82AB7E04
+				schema	default
+				follow	Some	domain	furore.dev
```
odisseus, line 1152: What is the meaning of this line?

odisseus (changes requested):
This pull request needs to be rebased so that we can see the test results in Github Actions.

.focus.fury
propensive: It's a change to the style of writing the layer path. It should make no difference to behavior.

layers/170FFB493AD839E66CD7F53CC22E808852B98777274C1F200EACE8BBA19AD952
propensive: It adds a dependency on another layer (also included). The `follow` line gives Fury a place to check for updates, though it's only partially implemented so far, so don't read too much into it yet.

PR #709, opened 2019-11-26
no pipelining

odisseus:


propensive:
I think pipelining is actually off by default anyway.

PR #714, opened 2019-11-27
Work in progress on nicer hint syntax

propensive:


propensive:
This looks like it will yield more fruitful results in the work in #982.

PR #715, opened 2019-11-27
Bootstrap with version 0.7.13

propensive:


odisseus:
You can also revert [the build script hack](https://github.com/propensive/fury/commit/195d39e6ddca02c9aa18c82a29ed10c8546cbcdb#diff-90829e76e906f1c73140c7ded7e1b268R22), because it should be unnecessary now.

PR #720, opened 2019-11-27
Support for running "interactive" (completions) tests

propensive:
These require `tmux` to be installed. They fake a command being typed
into a terminal in a particular shell, followed by a `tab` character,
which should invoke tab-completion for the appropriate shell.


etc/capture
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+tmux new-session -s fury -d $1
+case $1 in
+  'fish')
+    #tmux send-keys 'fishclear'
```
odisseus, line 6: Remove commented lines

odisseus (approved):

Issue #583, opened 2019-10-11
plugin module classes should be moved before use


propensive:
When a plugin is built from source using Bloop, the resultant classfiles reside in a temporary directory, created by Bloop. Downstream modules, however, are configured to use a plugin whose classes are stored in a fixed directory. It's not possible to specify the temporary directory in the downstream projects' configuration, so the only solution is to copy the files to the correct place immediately after compilation, so that the downstream module will know where to find them. This will also require that a plugin is treated as a separate subgraph.

PR #726, opened 2019-11-29
Treat compiler and plugin modules the same way as application modules

odisseus:
This pull request fixes the issue #583.


propensive (approved):
Is this all that was required to get the plugin to compile from source?

Issue #728, opened 2019-11-29
Generated Bloop configurations include unrelated modules in the classpath


odisseus:
The classpath that gets written to the Bloop configuration seems to depend not just on the actual dependencies of the target, but on all dependencies of the entire compilation object, which may include multiple targets.

As a result, the collection of available classes for each module depends on what else is compiled at the same time, even if there's no formal dependency.

propensive:
I think this was fixed. Please reopen if necessary, @odisseus.

PR #732, opened 2019-11-29
Limited classpath for targets

odisseus:
This pull request fixes the issue #728.


propensive (commented):
How confident are you that it works? And can we remove the logging?

odisseus:
My solution works, but it solves only half of the issue. This pull request won't be merged as it is; I'm going to prepare another one.

propensive (approved):
If this is producing working results, then I think it's an improvement on what we had before. Hopefully, if the performance isn't great, we can improve it later. I think there are a few opportunities for caching results. I will look at this in the future.

PR #741, opened 2019-12-03
Investigation of the way caching affects results of integration tests

odisseus:
| Commit                          | Fury cache | Integration |
| ---------------------------------------- | ------ | ------ |
| ffe3d654354c0301cbff4bb4a1423eefcad7fac6 | missed | passed |
| c5eba5fb370a1ace1801ea0f2811e8647ad32839 | missed | passed |
| a2e42cb6280d89018debdbe1ea94d007357d0030 | missed | passed |

### DIFF
--- a/dummy.txt
+++ b/dummy.txt
@@ -1,0 +1,1 @@
+Insignifficant change


odisseus:
Turns out the Fury cache [isn't saved because it became too large](https://github.com/propensive/fury/runs/330375232#step:35:3), so this PR is not going to trigger any cache hits. Regardless, integration tests pass for each commit when the cache is not working.

Issue #746, opened 2019-12-03
Layer extract into default location doesn't work on Mac OS


odisseus:
This is probably the cause behind failing Github Actions builds.
### Stack trace
```
$ fury standalone layer extract -f bootstrap/build.fury
Exception in thread "main" java.lang.ExceptionInInitializerError
	at fury.core.Log$.global(logging.scala:98)
	at fury.Main$.run(main.scala:89)
	at fury.Main$.main(main.scala:65)
	at fury.Main.main(main.scala)
Caused by: java.nio.file.FileAlreadyExistsException: /tmp
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:88)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:384)
	at java.nio.file.Files.createDirectory(Files.java:674)
	at java.nio.file.Files.createAndCheckIsDirectory(Files.java:781)
	at java.nio.file.Files.createDirectories(Files.java:727)
	at fury.io.Path.mkdir(path.scala:177)
	at fury.io.Path.extant(path.scala:104)
	at fury.model.Installation$.<init>(layout.scala:87)
	at fury.model.Installation$.<clinit>(layout.scala)
	... 4 more
```

odisseus:
It seems that the [NIO method](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-) [doesn't work if](https://stackoverflow.com/a/53956506) [one of the directories](https://github.com/propensive/fury/blob/538318a0c44d51557b91496c4241c535586916e5/src/model/layout.scala#L60) [in the path](https://github.com/propensive/fury/blob/538318a0c44d51557b91496c4241c535586916e5/src/model/layout.scala#L87) [is actually a symlink](https://apple.stackexchange.com/questions/1043/why-is-tmp-a-symlink-to-private-tmp).

PR #760, opened 2019-12-05
Layer extract into default location doesn't work on Mac OS  #746

ricardoffmartins:
Recover from exception, and use other method in case folders already exist.
https://github.com/propensive/fury/issues/746

ref: LayerExtract_MacOs

### DIFF
--- a/src/io/path.scala
+++ b/src/io/path.scala
@@ -28,7 +28,7 @@ import java.net.URI
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{FileVisitResult, Files, Paths, SimpleFileVisitor, StandardCopyOption, Path => JavaPath}
 import java.nio.file.StandardCopyOption._
-import java.io.{File => JavaFile, InputStream}
+import java.io.{InputStream, File => JavaFile}
 
 object Path {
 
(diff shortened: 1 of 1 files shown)


.focus.fury
```diff
@@ -1,2 +1,2 @@
 layerRef	3D11B9BCF50FBE70B34AAA410734DA6667123A7B5BB96AAFC8F6E187FC10B031
-path	/
+path	.
```
odisseus: @propensive, can you comment on this change?

src/io/path.scala
```diff
@@ -49,15 +49,24 @@ object Path {
...
+      Files.copy(file, targetPath.resolve(sourcePath.relativize(file)), REPLACE_EXISTING)
       FileVisitResult.CONTINUE
     }
   }
+
+  def createDirectories(path: JavaPath) = {
+    Try{
+      java.nio.file.Files.createDirectories(path)
+    }.recover{
+      case _:FileAlreadyExistsException =>
```
odisseus: Please make sure the new code is well formatted.

src/io/path.scala
```diff
@@ -49,15 +49,24 @@ object Path {
...
+        java.nio.file.Files.createDirectory(path)
```
odisseus: Can you explain why this works? The documentation [suggests](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectory-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-) that if method `createDirectories` threw a `FileAlreadyExistsException`, the method `createDirectory` will fail as well.

ricardoffmartins: createDirectories creates all directories needed (parent dirs) without thowing any exception in case if they already exists, except when we are creating a non Directory (a file) and it already exists.

createDirectory throws an exception if there is a directory in the pass with the same name, but not if there is a File with the same name.

This is what i understood from documentation.

.focus.fury
ricardoffmartins: fixed. Dont really know why it does that.

propensive: It should be `/`, not `.`. Layer paths are now written as `/foo/bar/baz` instead of `./foo/bar/baz`, with the root simply called `/`.

src/io/path.scala
propensive: I should come up with a style guide for code formatting...

src/io/path.scala
propensive: That doesn't sound quite right. If there's already a file there, creating a directory with the same name should never succeed...

propensive (commented):
@odisseus Can you have a closer look at the `createDirectory` vs `createDirectories` discrepancy, and merge if you're happy with it? I don't really understand what's going on! (And don't have time to investigate today.)

odisseus (approved):

Issue #759, opened 2019-12-05
develop a replacement model for schemas


propensive:
The model of using schemas, which are ad-hoc duplicates of the entire layer, does not lend itself well to maintainability. I would like to provide a more structured approach, whereby only the differences are stored, and "sets" of variations can be combined in useful ways.

This is far from a complete design. This issue exists as a reminder that it needs to be done.

PR #763, opened 2019-12-06
Renamed some short arguments

propensive:
Both files and directories are now specified with `-f`, which frees up `-d` to be used for dependencies.


src/build/build.scala
```diff
@@ -1,6 +1,6 @@
 /*
    ╔═══════════════════════════════════════════════════════════════════════════════════════════════════════════╗
-   ║ Fury, version 0.7.14. Copyright 2018-19 Jon Pretty, Propensive OÜ.                                         ║
```
odisseus: Could you separate this change into another pull request?

src/core/args.scala
```diff
@@ -59,9 +59,9 @@ object Args {
   val CompilerArg = CliParam[String]('c', 'compiler, "specify a compiler")
   val ClassArg = CliParam[String]('C', 'class, "specify a class name")
   val DefaultCompilerArg = CliParam[String]('c', 'compiler, "specify a default compiler")
-  val LinkArg = CliParam[String]('l', 'link, "specify a dependency link to another module")
-  val SourceArg = CliParam[String]('d', 'source, "specify a source directory")
-  val DirArg = CliParam[Path]('d', 'dir, "specify the new repository destination directory")
+  val DependencyArg = CliParam[String]('d', 'dependency, "specify a dependency on another module")
+  val SourceArg = CliParam[String]('f', 'source, "specify a source directory")
+  val DirArg = CliParam[Path]('f', 'dir, "specify the new repository destination directory")
```
odisseus, line 62: I think the previous flags were better. There is a difference between adding a source _file_ (which Fury does not support) and adding a source _directory_ that may or may not contain files at the moment.

propensive: I would like that, but I also want to have the invariant that for any entity type, the parameter corresponding to it should be a lower-case letter which is the first letter of the type of entity, where the reference is being made to an entity in the current layer. I would also like to use the same letter, but in upper-case, to refer to the same type of entity, but when it's external to the current layer.

propensive: We currently can't use `-s` for sources because `-s` is for schemas. I'm considering renaming "schemas" to "variants" and thus changing its parameter to `-v`, though. See #759.

odisseus: I'm not sure I understand your intent completely, but I support any kind of system which would help memorizing the arguments.

odisseus (approved):

odisseus:
The command `layer extract` accepts both `FileArg` and `DirArg`. Now that they both have the flag `-f`, that command is broken.

odisseus (changes requested):
The command `layer extract` [accepts](https://github.com/propensive/fury/blob/master/src/build/build.scala#L511) both `FileArg` and `DirArg`. Now that they both have the flag `-f`, that command is broken.

PR #766, opened 2019-12-06
Log everything to one log file per date

propensive:


odisseus:
Is the global log file going to be an addition or a replacement for the current log files? I think that the latter is a bad idea, as it would be hard to filter (or delete) the log messages that belong to a particular project, or a particular run of Fury.

propensive:
I've not decided one way or the other yet, or whether it should be configurable, but I don't want to have so many different log files created!

odisseus:
Why not? When reading the logs, I don't want to have to guess whether the messages I'm reading were produced by the action I'm analysing. If your log directory became cluttered, there is a [command](https://github.com/propensive/fury/blob/master/src/core/clean.scala#L43) to purge it.

propensive:
You shouldn't have to guess at anything. But the current setup, with a log file produced for every single invocation of `fury`, we end up with too many of them, with only slight differences in the timestamps assigned to them. Typically to find out what went wrong with some command you ran in the last minute, you might have to look at a dozen log files to find the right one. This change means you look at a single file, and maybe have to scroll a bit.

Issue #782, opened 2019-12-16
The layer select command goes crazy when the layer name is wrong


odisseus:
```
$ fury layer select -l platformff
$ cat .focus.fury 
layerRef    3D11B9BCF50FBE70B34AAA410734DA6667123A7B5BB96AAFC8F6E187FC10B031
path    /platform//platform
$ fury layer select -l platformff
$ cat .focus.fury 
layerRef    3D11B9BCF50FBE70B34AAA410734DA6667123A7B5BB96AAFC8F6E187FC10B031
path    /platform//platform//platform//platform
```

propensive:
Thanks, @ricardoffmartins!

PR #786, opened 2019-12-17
The layer select command goes crazy when the layer name is wrong #782

ricardoffmartins:
Create a function to verify if a layer exists.
New exception LayersFailure that is thrown when the given name doesn't exist.
Change the Root ImportPath from ImportPath("") to ImportPath("/")

ref: layerSelectCommand

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -498,16 +498,24 @@ object LayerCli {
     layout    <- cli.layout
     baseLayer <- Layer.base(layout)
     schema    <- baseLayer.mainSchema
-    cli       <- cli.hint(LayerArg, ImportPath("/") :: schema.importTree(layout, true).getOrElse(Nil))
+    cli       <- cli.hint(LayerArg,  schema.importTree(layout, true).getOrElse(Nil))
     call      <- cli.call()
-    _         <- schema.importTree(layout, true)
+    layers    <- schema.importTree(layout, true)
     relPath   <- call(LayerArg)
+    -         <- verifyLayers(relPath, layers)
     focus     <- Layer.readFocus(layout)
     newPath   <- ~focus.path.dereference(relPath)
     newFocus  <- ~focus.copy(path = newPath)
     _         <- Layer.saveFocus(newFocus, layout)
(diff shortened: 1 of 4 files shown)


src/build/build.scala
```diff
@@ -498,16 +498,24 @@ object LayerCli {
     layout    <- cli.layout
     baseLayer <- Layer.base(layout)
     schema    <- baseLayer.mainSchema
-    cli       <- cli.hint(LayerArg, ImportPath("/") :: schema.importTree(layout, true).getOrElse(Nil))
+    cli       <- cli.hint(LayerArg,  schema.importTree(layout, true).getOrElse(Nil))
```
propensive, line 501: Why was the `/` layer (the "root" layer) removed from the suggestions?

src/model/ids.scala
```diff
@@ -73,7 +73,7 @@ case class ModuleId(key: String) extends Key(msg"module")
 object ImportPath {
   implicit val msgShow: MsgShow[ImportPath] = ip => UserMsg(_.layer(ip.path))
   implicit val stringShow: StringShow[ImportPath] = _.path
-  val Root: ImportPath = ImportPath("")
+  val Root: ImportPath = ImportPath("/") //Can i change this?
```
propensive, line 76: I think so, though I would be interested to know why everything worked before, if the `Root` was incorrect...

propensive (approved):

src/build/build.scala
ricardoffmartins: It was added on ids.scala.

src/model/ids.scala
ricardoffmartins: Maybe because the Root "/" was being added inside the select function. and it would appear in completions.

ricardoffmartins:
@propensive 
there is also this 
https://github.com/ricardoffmartins/fury/blob/c33d3112acbcc2766479084953cfe78f0a4be484/src/model/ids.scala#L106

I didnt change this but is this correct?
(rawParts ++ xs) 
this looks like it could explain the weird path

PR #817, opened 2019-12-24
Fetch the repo when the latest version is needed

odisseus:
Fixes #816.

PR #875, opened 2020-01-01
Don't clone or checkout sources unnecessarily

propensive:


odisseus:
This is probably related to the pull request #817.

PR #882, opened 2020-01-03
Somewhat hacky attempt to improve IPFS download reliability

propensive:


propensive:
Looks like this doesn't really work as desired. It seems like it should be much less code.

propensive (commented):
Superseded by @odisseus's better fix.

PR #888, opened 2020-01-04
Run unit tests at same time as compilation of Fury

propensive:


propensive:
Apart from not working, it looks like it doesn't make things any faster anyway.

PR #916, opened 2020-01-10
Updated Bloop to version 1.4.0-RC1

odisseus:


propensive (approved):

propensive:
If you want to experiment a bit, it would be interesting to see if the `scala3` test continues to pass without the hack to run it twice in the (somewhat probable) case that it fails the first time.

propensive (approved):

PR #917, opened 2020-01-10
Migrated the layer format to version 5

odisseus:
There are no logic changes between the old layer and the new one. The large diff is caused by the changes in the format.


odisseus:
I forgot that the bootstrapping version doesn't support the new layer format.

PR #920, opened 2020-01-10
Added tests for parsing identifiers of various entities.

odisseus:


src/model-test/idtests.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.util.{Success, Try}
+
+object IdTests extends TestApp {
+
+  override def tests(): Unit = {
+
+    test("binary ids") {
+      BinaryId.unapply("foo.bar:baz-quux_2.12:1.2.3-RC4").isEmpty &&
+      BinaryId.unapply("foo.bar:baz-quux_2.12").isEmpty &&
+      BinaryId.unapply("baz-quux_2.12").isDefined
```
propensive: I think that this shouldn't actually be a valid ID. The purpose of the ID is not to encode a version number, and it should be actively discouraged. I'll need to update the code which infers a `BinaryId` from a `BinSpec`, though.

@odisseus Can you update the tests to reflect this? I'll update the method in a separate PR.

propensive (changes requested):

src/model-test/idtests.scala
odisseus: Ok

propensive:
Can this be marked "Ready for review" yet?

odisseus:
This pull request can be safely merged now, but I was going to add some more tests before merging.

PR #932, opened 2020-01-13
Clean Bloop caches before rebuilding

### LINKED ISSUE #883: old errors are displayed when running Fury in "watch" mode at the same time as Metals
Old error messages from previous failed compilations seem to get cached between build runs, and displayed even when compilation is successful. I suspect that this happens with modules whose compilation is deduplicated because Metals compiles them first.

odisseus:
https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/specification.md#clean-cache-request

This should fix #883.

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -279,7 +279,7 @@ object BuildCli {
     }
   }
 
-  private def repeater[T](sources: Set[Path])(f: Unit => T): Repeater[T] = new Repeater[T]{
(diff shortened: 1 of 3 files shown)


odisseus:
@propensive, can you check if this forks for you?

This branch adds some debugging output, which I'm going to remove before merging. However, the outdated error messages should not appear anymore, and rebuilding should be still fast when the change is small.

propensive:
Doesn't it make every iteration slower?

PR #936, opened 2020-01-15
Updated Bloop to version 1.4.0-RC1+9-e94b10c3

### LINKED ISSUE #883: old errors are displayed when running Fury in "watch" mode at the same time as Metals

odisseus:
Fixes #883.


odisseus:
Nope, it doesn't.

PR #942, opened 2020-01-16
Write BSP trace to file

odisseus:
The BSP trace is now enabled by a configuration setting:
```
fury config set --bsp-trace true
```

The default value is **false**.

The logs are stored inside the work directory, in files named like `.fury/logs/2020-01-17T14:54:02.023.log`


src/build/build.scala
```diff
@@ -432,14 +434,15 @@ object BuildCli {
                                 pipelining: Boolean,
                                 reporter: Reporter,
                                 theme: Theme,
-                                https: Boolean)
+                                https: Boolean,
+                                bspTrace: Option[Path])
```
propensive: See my comment on the issue, but I think that logging to the global log should be the only option. My reasoning is that if we follow this example, we could end up offering different log files for every external program we integrate with, and I don't want to go down that route. We already have a global log, and we can use that to interleave BSP output with other things that Fury is doing, in chronological order.

src/compile/compile.scala
```diff
@@ -145,7 +154,8 @@ object BloopServer {
       Connection(proxy, client, thread)
     }
 
-  def borrow[T](dir: Path, multiplexer: Multiplexer[ModuleRef, CompileEvent], compilation: Compilation, targetId: TargetId, layout: Layout)(fn: Connection => T)(implicit log: Log): Try[T] = {
+  def borrow[T](dir: Path, multiplexer: Multiplexer[ModuleRef, CompileEvent], compilation: Compilation, targetId: TargetId, layout: Layout, bspTrace: Option[Path])(fn: Connection => T)(implicit log: Log): Try[T] = {
+    //FIXME a connection will inherit the trace destination (or lack thereof) from its first usage
```
propensive: This would be fixed by using the global log.

src/core/args.scala
```diff
@@ -95,6 +95,7 @@ object Args {
   val ServiceArg = CliParam[String]('S', 'service, "specify the default remote layer service")
   val TargetArg = CliParam[String]('T', 'target, "target file/directory")
   val TransformArg = CliParam[Unit]('t', 'transform, "transform the option into the parameters following --")
+  val TraceArg = CliParam[Path]('t', Symbol("bsp-trace"), "Write the BSP communication trace to a file")
```
propensive: Make it a `CliParam[Unit]` to make it a flag.

propensive (changes requested):
I think we should simplify this to use the global log.

src/build/build.scala
odisseus: The BSP trace is different from logs generated by Fury. I agree that it is convenient to see the exact chronological order of all events that happened, but intermingling BSP traces with the logs would make both very hard to read.

The BSP server is quite different from other external programs, because it is by far the most tightly integrated and has the most complex interaction with Fury. I think it is reasonable to make an exception and write its traces into a separate file, because these aren't just another kind of Fury logs. It won't make a big impact on the typical user, because by default the traces would not be written at all. Normal logs related to the BSP interaction would of course still go to the global log file.

src/compile/compile.scala
odisseus: I resolved this problem by making the configuration global.

propensive (commented):
Can we do it without the new dependency?

src/core/args.scala
odisseus: This parameter has to be boolean now, so that the user can set the configuration property to true or false.

odisseus:
Looks like it is quite feasible. The easiest way would be copying the source code of [Apache's `TeeOutputStream`](https://github.com/apache/commons-io/blob/b8f638fa691dc600248b9d810927d0491a46bc0f/src/main/java/org/apache/commons/io/output/TeeOutputStream.java).

propensive:
I think we can probably implement the relevant parts of the `TeeOutputStream` in just a few lines of Scala. I'll have a go later, then I think this is ready to merge.

odisseus:
Done

PR #943, opened 2020-01-18
Only run Scala 3 once

propensive:
This should test whether the latest version of Bloop works now.


propensive:
This has failed to run correctly one time out of six. For now, that's unreliable enough for me to keep the hack in place.

PR #945, opened 2020-01-19
Shorter `TeeOutputStream` supporting more than two `OutputStream`s

propensive:


src/utils/teeoutputstream.scala
```diff
@@ -14,44 +14,23 @@
...
-  }
-
+class TeeOutputStream(streams: OutputStream*) extends OutputStream {
+  override def write(b: Array[Byte], off: Int, len: Int): Unit =
+    synchronized(streams.foreach(_.write(b, off, len)))
+  
+  override def write(b: Array[Byte]): Unit = synchronized(streams.foreach(_.write(b)))
+  override def write(b: Int): Unit = synchronized(streams.foreach(_.write(b)))
+  override def flush(): Unit = streams.foreach(_.flush())
+  override def close(): Unit = streams.map { stream => Try(stream.close()) }
```
odisseus, line 35: This method swallows any exceptions that might happen when the streams are closed.

PR #949, opened 2020-01-20
Permanent BSP connections

odisseus:
Problems:

- [x] Shutdown method is never called at the moment
    - [x] Shutdown method is never called in standalone mode
- [x] The console output (such as diagnostic messages) works only the first time when a connection is used
    - [x] The connection is bound to a multiplexer, but said multiplexer becomes unusable after the first compilation result
- [x] Multiplexer instances are never deleted
    - Loooks like they are deleted after all
- [ ] BSP trace logs are broken
    - This always happens in integration tests, because the destination directory is deleted.
    - Sometimes it happens even if the destination directory remains in place.


propensive:
It looks like two of the integration test failures are just removal of the repeated BSP connection logs, but I can't explain the `scala3` test failure.

src/build/build.scala
```diff
@@ -201,6 +205,9 @@ object BuildCli {
                                  |
                                  |${builds}
                                  |
+                                 |Active BSP connections:
```
propensive: I'm going to create an issue to create a separate command to show this information.

propensive: (but it's fine here, for now)

src/core/lifecycle.scala
```diff
@@ -20,9 +20,16 @@ import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.annotation.tailrec
 import scala.collection.mutable.HashSet
+import scala.concurrent.{Await, Future, Promise}
 import scala.util.Try
 
 object Lifecycle {
+
```
propensive, line 27: I'm interested to see where this is going, but I'll wait until it's done! Do you think there's an opportunity to generalize several different things we might want to shut down using this?

odisseus: I hope this will eventually happen, but I don't want to touch in this PR anything that's not related to the BSP connections. Besides, I'm still not sure how the generalized solution should look.

src/build/build.scala
odisseus: The "about" command could have further modifiers, like "about tasks", "about connections" etc. But that's out of scope for this pull request. I may revert this particular change before merging, but for now I need it for the purpose of debugging.

PR #964, opened 2020-01-22
Use `Try` for `~` instead of assuming `Success`

propensive:


odisseus:
I thought hiding exceptions in the `~` operator was a design decision...

propensive:
It was motivated by more convenient syntax for things we didn't expect to fail (which we wanted to include in a for-comprehension). It turns out that some of them actually fail sometimes. This change means that we actually catch those exceptions in the for-comprehension now, instead of elsewhere (or not). It should, very occasionally, give us better errors.

PR #970, opened 2020-01-23
Task infrastructure

propensive:


layers/ABDC3E79197F626461DADE8619BAA995647A906420E5C3B42C42BBE0A383F13B
```diff
@@ -0,0 +1,345 @@
+version	5
```
odisseus, line 1: How did you get the integration to work with the new layer format?

propensive: I think there's a mistake. This was just a quick check-in of all my WIP before I had to leave, so it's not anywhere near ready yet, and shouldn't compile... which leads me to conclude that `.focus.fury` is not pointing to the right layer file (even if `.fury.conf` is), so the build isn't using the right layer file, so it's not attempting to compile the `tasks.scala` file I checked in.

PR #972, opened 2020-01-24
Use `procname.c` to set the process name to `fury`

propensive:
Also, do C compilation during install, not during package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ publish: dist/install.sh
 opt:
 	mkdir -p opt
 
+dist/bundle/bin/procname.c:
+	cp etc/procname.c $@
+
 opt/fury-$(FURYSTABLE).sh: opt
 	curl -C - -s -o $@ "http://downloads.furore.dev/fury-$(FURYSTABLE).sh"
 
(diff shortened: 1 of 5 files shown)


propensive:
This makes `killall fury` work, when necessary.

odisseus:
On my computer it works even better than expected:
```
$ jps -l
6596 bloop.Server
6468 com.facebook.nailgun.NGServer
6711 sun.tools.jps.Jps
$ killall fury
$ jps -l
6749 sun.tools.jps.Jps
```

odisseus (changes requested):
The C compilation doesn't work on Mac OS:
```
dist/install.sh 

# ...

/Users/user919720/.fury/bin/procname.c:1:10: fatal error: 'sys/prctl.h' file not found
#include <sys/prctl.h>
         ^~~~~~~~~~~~~
1 error generated.
 * Native Nailgun client compilation failed
 * Python Nailgun client will be used
```

propensive:
I think the compilation failure is fine. It has an error, but should gracefully degrade to the old behavior. I'll send all output to `/dev/null` though.

PR #976, opened 2020-01-24
Cancel BSP requests

odisseus:


odisseus:
This approach seems to be a dead end. Java's completable futures work exactly like promises in Scala: they can be completed early, but completion doesn't automatically abort the task associated with it.

PR #982, opened 2020-01-27
Move `cli` and `log` params to `...Cli` objects

propensive:


src/build/build.scala
```diff
@@ -72,9 +72,9 @@ object ConfigCli {
   } yield log.await()
 }
 
-object AboutCli {
+case class AboutCli(cli: Cli)(implicit log: Log) {
 
-  private def resources: String = {
+  private def resourcesString: String = {
```
odisseus, line 77: The type of this method already says it returns a string — why repeat that in its name?

propensive: Only because there was already a method called `resources`, at least when I wrote it. I may have deleted it later. I agree that it's generally undesirable to include the type in the name.

PR #984, opened 2020-01-27
Better management of multiplexers and command line sessions

### LINKED ISSUE #983: graph is not updated until compilation finishes
The problem was introduced in #949, and is probably related to the multiplexer becoming a `var`. The update messages may be being sent to the wrong multiplexer now.

odisseus:
The class `FuryBuildClient` now has no mutable references — all the state related to switching of command line clients is now stored in the global objects `BloopServer` and `Lifecycle`, which serve as a kind of "multiplexer service" for the build clients.

The top of the compilation graph is distorted by the logs of BSP initialization, but I think a better way to fix this would be redrawing the whole graph after those lines of logs.

Fixes #983.

### DIFF
--- a/src/compile/compile.scala
+++ b/src/compile/compile.scala
@@ -70,13 +70,15 @@ object BloopServer extends Lifecycle.Shutdown with Lifecycle.ResourceHolder {
     }
   }
 
+  def subscribers(bspClient: FuryBuildClient): List[Session] = usages.synchronized {
+    usages.collect { case (session, connection) if connection.client == bspClient => session }.to[List]
+  }
+
   override def acquire(session: Session, connection: Connection): Unit = usages.synchronized {
     usages += session -> connection
-    connection.client.session = session
   }
 
   override def release(session: Session): Unit = usages.synchronized {
-    //TODO consider unsetting reference to session in connection.client
     usages -= session
   }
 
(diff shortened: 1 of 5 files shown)


src/core/lifecycle.scala
```diff
@@ -28,29 +31,47 @@ object Lifecycle {
...
   
-  val bloopServer = Promise[Shutdown]
+  val bloopServer = Promise[Shutdown with ResourceHolder]
 
   case class Session(cli: Cli, thread: Thread) {
     val pid = cli.pid
     val started: Long = System.currentTimeMillis
+    private var _multiplexer:  Multiplexer[ModuleRef, CompileEvent] = new Multiplexer(List.empty)
+    def multiplexer = synchronized{ _multiplexer }
+    def multiplexer_=(m: Multiplexer[ModuleRef, CompileEvent]) = synchronized{ _multiplexer = m }
```
propensive, line 48: Where did you learn about this syntax? ;) I thought nobody knew about it any more!

src/compile/compile.scala
```diff
@@ -50,12 +51,14 @@ import scala.util.control.NonFatal
 
 trait FuryBspServer extends BuildServer with ScalaBuildServer
 
-object BloopServer extends Lifecycle.Shutdown {
+object BloopServer extends Lifecycle.Shutdown with Lifecycle.ResourceHolder {
```
propensive, line 54: I'm unconvinced that inheritance of `Shutdown` and `ResourceHolder` provides much benefit, but it's not worth removing if what we've got is working.

propensive (approved):
This looks like an improvement, but it's difficult to evaluate just by reading the code, so I think we'll discover whether it's really better with some real-world usage.

src/compile/compile.scala
odisseus: Surely these traits must be refactored in the future — I just don't know yet how exactly. Maybe it will become more clear after the cancellation of tasks is implemented.

Issue #991, opened 2020-01-29
NoSuchElementException when cloning a layer from IPFS


odisseus:
I have seen this on two different computers, but it doesn't happen every time. The exception seems to happen only during the first few attempts to clone the layer, and only if the IPFS daemon hasn't been started since computer reboot. If the daemon was started some time ago but was killed later, this exception doesn't happen.

### Stack trace
```
java.util.NoSuchElementException: None.get
    at scala.None$.get(Option.scala:366)
    at scala.None$.get(Option.scala:364)
    at fury.core.Ipfs$.fury$core$Ipfs$$install$1(ipfs.scala:108)
    at fury.core.Ipfs$$anonfun$daemon$10.$anonfun$applyOrElse$1(ipfs.scala:124)
    at scala.util.Failure.orElse(Try.scala:224)
    at fury.core.Ipfs$$anonfun$daemon$10.applyOrElse(ipfs.scala:124)
    at fury.core.Ipfs$$anonfun$daemon$10.applyOrElse(ipfs.scala:119)
    at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:38)
    at scala.util.Failure.recoverWith(Try.scala:236)
    at fury.core.Ipfs$.daemon(ipfs.scala:119)
    at fury.core.Layer$.$anonfun$loadFromIpfs$1(layer.scala:48)
    at fury.model.Installation$.$anonfun$tmpDir$1(layout.scala:136)
    at fury.model.Installation$.tmpFile(layout.scala:142)
    at fury.model.Installation$.tmpDir(layout.scala:134)
    at fury.core.Layer$.loadFromIpfs(layer.scala:47)
    at fury.core.Layer$.load(layer.scala:116)
    at fury.LayerCli$.$anonfun$clone$7(build.scala:536)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.LayerCli$.$anonfun$clone$6(build.scala:535)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.LayerCli$.$anonfun$clone$5(build.scala:534)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.LayerCli$.$anonfun$clone$4(build.scala:533)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.LayerCli$.$anonfun$clone$3(build.scala:532)
    at scala.util.Success.flatMap(Try.scala:251)
[...]

PR #992, opened 2020-01-29
Better exception when uname returns unexpected values

odisseus:
This will help to find the cause behind the issue #991. I suspect that uname returns bad data on some occasions.


propensive (commented):
I thought that the behavior was already to just report an error when attempting to install IPFS if it couldn't determine the OS details...

PR #1004, opened 2020-01-30
Better zsh prompt, consistent with context string

propensive:


odisseus:
I guess the integration tests need to be updated as well.

propensive:
Yes. I think there's a separate bug which is showing the prompt in raw output during removal of resources, when it shouldn't, but that's something I'll fix later.

PR #1007, opened 2020-01-30
Don't skip compilation for executable modules that have no own sources

### LINKED ISSUE #998: An application module is not executed if it has no source directories
A workaround is creating an empty source directory. This is sufficient to trigger a BSP compile request and subsequent execution of the main class.

odisseus:
Fixes #998.


test/passing/empty-application/script
```diff
@@ -8,7 +8,14 @@ fury project add -n hello-world
 fury module add -n library -c scala/compiler
 fury source add -d src
 mkdir -p src
-echo 'object HelloWorld { println("Hello world!") }' > src/hello.scala
+cat << EOF >  src/hello.scala
```
propensive, line 11: For future reference, it's probably easier just to create this file in the `empty-application/src` directory, and check it in.

propensive (approved):

PR #1033, opened 2020-02-04
Create the log file again if it has been deleted

### LINKED ISSUE #1028: Fury clean breaks BSP trace
I think this happens because the logs directory is deleted, which breaks the output stream.

odisseus:
I have checked manually that this works on Fedora 31 on an ext4 file system. I am going to add a unit test in a separate pull request.

Fixes #1028.


src/io/recoveringfileoutputstream.scala
```diff
@@ -0,0 +1,44 @@
...
+   ║ See the License for the specific language governing permissions and limitations under the License.        ║
+   ╚═══════════════════════════════════════════════════════════════════════════════════════════════════════════╝
+*/
+package fury.io
+
+import java.io._
+
+class RecoveringFileOutputStream(path: Path, append: Boolean) extends OutputStream {
+
+  private def createStream(): FileOutputStream = {
```
propensive: Should be `private[this]`

propensive (changes requested):
I would prefer to go with #1038 as a solution to this.

src/io/recoveringfileoutputstream.scala
odisseus: Ok

PR #1035, opened 2020-02-04
Aliases with arguments

propensive:


odisseus:
This pull request breaks most console commands; they just exit immediately with code 1.

propensive:
Any idea why? It looks like they shouldn't cause any issues... Maybe the change to `alias` means that the layer file can't be read?

propensive:
Actually, that's probably it. I've added a default value (`Nil`) for the application arguments.

odisseus (approved):

PR #1041, opened 2020-02-06
Don't fail layer operations if BSP config generation fails

### LINKED ISSUE #1040: Generation of the BSP config fails if Fury is not present in the classpath
The code that generates the configuration file for using Fury as a BSP server assumes that `which fury` will always print a valid path and exit with code 0. However, this assumption doesn't hold on a clean system, on which Fury has never been installed before. Due to BSP configuration being generated by the `layer extract` command (and also `layer init`), installation on a clean system will always fail.

This issue was introduced by PR #996 and caused failures in PR #1032.

The hack introduced in commit 7273f7b6618f39d763c42142e191e99b947ecd2c fixes the installation process, but does not solve the issue completely.

odisseus:
Config generation would fail if Fury is not present in `$PATH`. However, this should not prevent using Fury directly.

This pull request fixes #1040, but it has an unpleasant side effect: when Fury is installed from source in a fresh environment, the BSP server configuration will not be generated for the Fury project itself.

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -459,7 +459,7 @@ case class LayerCli(cli: Cli)(implicit log: Log) {
     layout <- cli.newLayout
     call   <- cli.call()
     _      <- Layer.init(layout)
-    _      <- Bsp.createConfig(layout)
+    _      =  Bsp.createConfig(layout)
   } yield log.await()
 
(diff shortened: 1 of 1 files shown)


propensive:
Also have a look at PR #1042.

PR #1042, opened 2020-02-06
Use `fury.home` property to determine location of `fury` command

propensive:


odisseus:
The makefile hack from 313eab4ea644240040fd973463efa53e9c3b4866, as well as the entire commit ee53d531010760659065295528f9523e1d6a0d4f, can be reverted in this pull request.

odisseus:
This seems to be a good solution for all use cases except one: when Fury is installed from sources in a fresh environment, the generated BSP config will contain not the path where Fury is going to be installed (which is unknown until the very end of the installation), but the path where the bootstrapping version is located.

odisseus (approved):

PR #1045, opened 2020-02-07
Use a different port for Bootstrapping Fury version

propensive:


odisseus:
Try `kill $(shell lsof -i:$(FURY_PORT) -t)`

odisseus:
The configuration for Github Actions should probably be updated as well.
https://github.com/propensive/fury/blob/d357c5894349d50a1ab4c8f2eac83930ea8deae0/.github/workflows/main.yml#L23
https://github.com/propensive/fury/blob/d357c5894349d50a1ab4c8f2eac83930ea8deae0/.github/workflows/main.yml#L56

Issue #1044, opened 2020-02-07
handle SIGTERM signals to shutdown Fury


propensive:
Fury should handle a request to die using `Runtime.addShutdownHook()`.

PR #1050, opened 2020-02-10
Shutdown hook

odisseus:
Killing the NGServer process now [shuts down](https://github.com/build-server-protocol/build-server-protocol/blob/master/docs/specification.md#shutdown-build-request) the BSP connections.

Related to #1044.


propensive:
Note that `NGServer` should now be called `Fury` since #1106.

PR #1064, opened 2020-02-13
Test for writing to a directory as if it were a file

odisseus:


src/io/path.scala
```diff
@@ -162,11 +162,12 @@ case class Path(input: String) {
...
 
-  def writeSync(content: String, append: Boolean = false): Try[Unit] = Try {
-    val writer = new java.io.BufferedWriter(new java.io.FileWriter(javaPath.toFile, append))
-    writer.write(content)
-    Success(writer.close())
-  }.transform(identity, e => Failure(FileWriteError(this, e)))
+  def writeSync(content: String, append: Boolean = false): Try[Unit] = {
+    val writer = Try { new java.io.BufferedWriter(new java.io.FileWriter(javaPath.toFile, append)) }
+    val result = writer.flatMap{ w => Try(w.write(content)) }
+    writer.foreach(_.close())
```
propensive, line 168: Replace with something like,
```
writer.flatMap { w => w.close(); result }
```

propensive (changes requested):

src/io/path.scala
odisseus: I'll try implementing a general solution in another pull request.

PR #1082, opened 2020-02-19
Work in progress new install script

propensive:


.mergify.yml
```diff
@@ -1,9 +0,0 @@
-pull_request_rules:
```
odisseus, line 1: Why?

Issue #1077, opened 2020-02-17
Unrecognized targets in BSP server mode


odisseus:
## Steps to reproduce
1. Launch Fury as a BSP server, e. g. using IntelliJ as a client.
2. Request compilation of one of the intermediate modules e. g. `fury/model`.
3. Request compilation of `fury/frontend`

**Expected behaviour:** Compilation is completed without errors both times
**Actual behaviour:** Lots of stack traces are printed during the second compilation, but it still succeeds

It seems that exceptions are thrown for every dependency of `fury/frontend` which is not a dependency of `fury/model`.

## Example stack trace
```
Feb 17, 2020 7:58:48 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleNotification
WARNING: Notification threw an exception: {
"jsonrpc": "2.0",
"method": "build/taskStart",
"params": {
"taskId": {
"id": "23"
},
"eventTime": 1581965928620,
"message": "Compiling default_eucalyptus_core (1 Scala source)",
"dataKind": "compile-task",
"data": {
"target": {
"uri": "file:/home/virtuslab/devel/fury/.fury/work/default_eucalyptus_core/?id\u003ddefault_eucalyptus_core"
}
}
}
}
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify(GenericEndpoint.java:152)
at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification(RemoteEndpoint.java:220)
at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:187)
at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:192)
at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:99)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

Issue #987, opened 2020-01-28
Live updates from a BSP connection are sent only to one of the clients


odisseus:
## Steps to reproduce
1. Run `fury clean` so that the following tasks take noticeable time.
2. Run `fury build` and wait until the build graph appears.
3. Run `fury build` in another terminal.

**Expected behaviour:** in both terminals, graphs are updated as the build progresses
**Actual behaviour:** While the graph in the first terminal is updated, the second terminal displays an empty graph. After the first terminal reports that the build has been completed, there is a delay of a couple seconds, and the second graph gets filled very quickly.

propensive:
Check that the second graph is processing a `flush()` to its `stdout` every time.

PR #1085, opened 2020-02-19
Untangle Compilation object from BSP client

odisseus:
This pull request is related to issues #987 and #1077.

### DIFF
--- a/src/compile/compile.scala
+++ b/src/compile/compile.scala
@@ -300,7 +300,7 @@ object Compilation {
 
 class FuryBuildClient(targetId: TargetId, layout: Layout) extends BuildClient {
 
-  private def broadcast(event: CompileEvent): Unit = {
+  def broadcast(event: CompileEvent): Unit = {
     val ref = event match {
       case e: ModuleCompileEvent => e.ref
       case _ => targetId.ref
@@ -641,13 +641,15 @@ case class Compilation(graph: Target.Graph,
         soi.setClassDirectory(permanentClassesDir.javaFile.toURI.toString)
       }
 
-      result.get
+      (result.get, conn.client)
     }.map {
-      case compileResult if compileResult.isSuccessful && target.kind.needsExecution =>
+      case (compileResult, client) if compileResult.isSuccessful && target.kind.needsExecution =>
         val classDirectories = compileResult.classDirectories
+        client.broadcast(StartRun(target.ref))
         val exitCode = run(target, classDirectories, layout, globalPolicy, args, noSecurity)
+        client.broadcast(StopRun(target.ref))
         compileResult.copy(exitCode = Some(exitCode))
-      case otherResult =>
+      case (otherResult, _) =>
         otherResult
     }
   }
--- a/src/compile/compile.scala
+++ b/src/compile/compile.scala
@@ -213,6 +213,8 @@ object Compilation {
 
   private val requestOrigins: collection.mutable.Map[String, Compilation] = TrieMap()
 
+  def findBy(targetId: TargetId): Iterable[Compilation] = requestOrigins.values.filter(_.target.id == targetId)
+
   def mkCompilation(schema: Schema,
                     ref: ModuleRef,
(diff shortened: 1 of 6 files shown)


src/compile/compile.scala
```diff
@@ -211,6 +211,8 @@ object Compilation {
 
   private val compilationCache: collection.mutable.Map[Path, Future[Try[Compilation]]] = TrieMap()
 
+  val requestOrigins: collection.mutable.Map[String, Compilation] = TrieMap()
```
propensive: Can we make this private?

src/compile/compile.scala
```diff
@@ -146,7 +146,7 @@ object BloopServer extends Lifecycle.Shutdown with Lifecycle.ResourceHolder {
 
       thread.start()
       
-      val client = new FuryBuildClient(compilation, targetId, layout)
+      val client = new FuryBuildClient(targetId, layout)
```
propensive, line 148: Great!

src/compile/compile.scala
```diff
@@ -285,21 +286,26 @@ object Compilation {
   }
 
   private val originIdCounter = new AtomicInteger(1)
-  def nextOriginId()(implicit log: Log): String = {
-    s"${log.pid.pid}-${originIdCounter.getAndIncrement}"
+
+  def nextOriginId(compilation: Compilation)(implicit log: Log): String = {
+    val originId = s"${log.pid.pid}-${originIdCounter.getAndIncrement}"
```
propensive: Is there a reason why we're using a stringly-typed index for the origin ID? Can you create an `OriginId` type in `ids.scala`?

src/compile/compile.scala
```diff
@@ -285,21 +286,26 @@ object Compilation {
...
 
-  private def broadcast(event: CompileEvent): Unit = {
+  def broadcast(event: CompileEvent): Unit = {
     val ref = event match {
       case e: ModuleCompileEvent => e.ref
       case _ => targetId.ref
     }
-    BloopServer.subscribers(this).map(_.multiplexer).foreach(_(ref) = event)
+    BloopServer.subscribers(this).map(_.multiplexer)
+      .filter(_.contains(ref)).foreach(_(ref) = event)
```
propensive, line 307: It's probably fine, but can you check whether `contains` (on whatever `multiplexer` is) is constant time, i.e. that it's not linear for every single event we receive (and hence _O(nm)_).

src/compile/compile.scala
```diff
@@ -389,22 +396,32 @@ ${'|'} ${highlightedLine}
...
       val targetId = getCompileTargetId(params.getData)
       val ref = targetId.ref
-      broadcast(StopCompile(ref, params.getStatus == StatusCode.OK))
-      if(!compilation.targets(ref).kind.needsExecution) broadcast(StopRun(ref))
-      else broadcast(StartRun(ref))
+      broadcast( StopCompile(ref, params.getStatus == StatusCode.OK))
+      val compilations = Compilation.requestOrigins.values.filter(_.target.id == targetId)
+      for {
+        compilation <- compilations
+      } yield {
```
propensive: I think, `compilations.foreach { ... }` makes it clearer that this returns `Unit`.

src/compile/compile.scala
propensive: Ok, I saw your implementation below. It defers to `keys` in the multiplexer which is a `List`, so it's linear. Can we make `keys` a `Set`?

propensive (changes requested):

src/compile/compile.scala
odisseus: Good point

src/compile/compile.scala
odisseus: Refactored

src/compile/compile.scala
odisseus: Refactored

propensive (approved):

PR #1087, opened 2020-02-20
Refactored conversion of build target URIs to internal target identifiers, and added a test.

odisseus:


propensive (changes requested):
1. Avoid structural subtyping and `asInstanceOf` by passing the result of `.getData` directly to the method.
2. Remove the schema (always `default_`) from the target IDs.

PR #1098, opened 2020-02-26
Quoting, quoting everywhere

odisseus:


propensive (approved):
I'm working on a new Makefile and install script in the `new-install-script` branch, which adheres to draconian quoting of everything, but this is all good.

PR #1103, opened 2020-02-27
Fixed IPFS installation

odisseus:


src/core/external.scala
```diff
@@ -105,7 +105,8 @@ object Ipfs {
       case e: RuntimeException if e.getMessage.contains("Couldn't connect to IPFS daemon") =>
         log.infoWhen(!quiet)(msg"Couldn't connect to IPFS daemon")
         for {
-          ipfs <- IpfsSoftware.installedPath(env, quiet)
+          _    <- IpfsSoftware.installedPath(env, quiet)
+          ipfs =  IpfsSoftware.managedPath
```
propensive: This looks wrong... `installedPath` will install `ipfs` only if it's required, then the second line will use the path which assumes IPFS is installed.

src/core/external.scala
```diff
@@ -143,9 +144,9 @@ abstract class Installable(name: ExecName) extends Software(name) {
       _   <- ~log.infoWhen(!quiet)(msg"Downloading $bin...")
       in  <- Http.requestStream(bin, Map[String, String](), "GET", Set())
       _   <- TarGz.extract(in, base)
-      _   <- ~managedPath.setExecutable(true)
+      _   <- managedPath.setExecutable(true)
```
propensive, line 147: With the fix for `TarGz.extract`, this line should be redundant anyway.

src/io/path.scala
```diff
@@ -122,7 +122,7 @@ case class Path(input: String) {
     this
   }
 
-  def setExecutable(exec: Boolean): Try[Unit] = Try(javaFile.setExecutable(exec))
+  def setExecutable(exec: Boolean): Try[Unit] = Try(javaFile.setExecutable(exec)).collect{ case true => () }
```
propensive: Maybe we could transform the exception type into something better?

propensive (commented):
So, does this make things work? It seems surprising... (and even if it does work, I doubt it will use an existing installation of `ipfs` when it should).

src/core/external.scala
odisseus: Good point; this must be the cause behind the timeouts I'm getting.

src/core/external.scala
odisseus: I'm not so sure. In an arbitrary distribution archive, the executable files may or may not have the correct permissions, whereas we need to make sure they are set once these files are extracted.

src/io/path.scala
odisseus: Done

PR #1107, opened 2020-02-28
Allow users to select `java/compiler` as the compiler

propensive:


test/passing/set-java-compiler/script
```diff
@@ -0,0 +1,4 @@
+fury layer init
+fury project add -n test
+fury module add -c java-compiler -n test
```
odisseus: Is this just a typo or is it a new "implied" compiler module?

propensive: It's a typo. Good catch. I wasn't able to test this when I was writing it, so I didn't notice it...

Issue #900, opened 2020-01-07
generalize config menu


propensive:
We shouldn't need to have a separate flag for every setting. It should be commands all the way down, e.g.
```
fury config theme # shows the current theme
fury config timestamps on # set the timestamps to on
```

This would leave much more scope for additional configuration options.

PR #1108, opened 2020-02-28
Try using HTTP gateways if IPFS get fails

### LINKED ISSUE #1102: Try to resolve layers using HTTP gateways if IPFS operation fails
API example: 
https://gateway.pinata.cloud/api/v0/get?arg=Qmf24DLiVqXfpQtzWzmN8cRuYYLmNACKeA58ccjLqD2557&archive=true

odisseus:
I have checked manually that this works.

Closes #1102.

### DIFF
--- a/src/core/external.scala
+++ b/src/core/external.scala
@@ -32,6 +32,11 @@ object Ipfs {
   import io.ipfs.api._
   import io.ipfs.multihash.Multihash
 
+  private val knownGateways = List(
+      "ipfs.io",
+      "gateway.pinata.cloud"
+  )
+
   case class IpfsApi(api: IPFS){
     def add(path: Path): Try[IpfsRef] = Try {
       val file = new NamedStreamable.FileWrapper(path.javaFile)
@@ -40,7 +45,18 @@ object Ipfs {
       IpfsRef(top.hash.toBase58)
     }
 
-    def get(ref: IpfsRef, path: Path): Try[Path] = getFile(Multihash.fromBase58(ref.key), path)
+    def get(ref: IpfsRef, path: Path)(implicit log: Log): Try[Path] = {
+      val hash = Multihash.fromBase58(ref.key)
+      getFile(hash, path).recoverWith { case e =>
+        log.warn("Could not resolve the hash using IPFS daemon")
+        knownGateways.foldLeft[Try[Path]](Failure(e)){
+          case (res@Success(_), _) =>res
+          case (Failure(_), gateway) =>
+            log.warn(s"Could not resolve the hash using $gateway")
+            getFileFromGateway(hash, path)(gateway)
+        }
+      }
+    }
 
     def id(): Try[IpfsId] = Try {
       val idInfo = api.id()
@@ -56,12 +72,23 @@ object Ipfs {
       )
     }
 
-    private def getFile(hash: Multihash, path: Path): Try[Path] = for {
+    private[this] def getFile(hash: Multihash, path: Path): Try[Path] = for {
       _    <- path.mkParents()
       data =  api.get(hash)
       in   =  new ByteArrayInputStream(data)
       _    =  TarGz.untar(in, path)
     } yield path.childPaths.head
+
+    private[this] def getFileFromGateway(hash: Multihash, path: Path)(gateway: String)(implicit log: Log): Try[Path] = {
+      log.info(s"Accessing $gateway to retrieve $hash")
+      val query = Query & ("arg" -> hash.toBase58) & ("archive" -> "true")
+      for {
+        _    <- path.mkParents()
+        data <- Http.get(Https(Path(gateway) / "api" / "v0" / s"get", query), Set.empty)
+        in   =  new ByteArrayInputStream(data)
+        _    =  TarGz.untar(in, path)
+      } yield path.childPaths.head
+    }
 
   }
 


src/core/external.scala
```diff
@@ -32,6 +32,11 @@ object Ipfs {
   import io.ipfs.api._
   import io.ipfs.multihash.Multihash
 
+  private val knownGateways = List(
```
propensive, line 35: Good to build in some sort of choice. I'd like to make this configurable, but only once I've done a small overhaul of the configuration interface. (Issue #900)

src/core/external.scala
```diff
@@ -40,7 +45,15 @@ object Ipfs {
       IpfsRef(top.hash.toBase58)
     }
 
-    def get(ref: IpfsRef, path: Path): Try[Path] = getFile(Multihash.fromBase58(ref.key), path)
+    def get(ref: IpfsRef, path: Path)(implicit log: Log): Try[Path] = {
+      val hash = Multihash.fromBase58(ref.key)
+      getFile(hash, path).recoverWith { case e =>
+        log.warn(s"Could not resolve the hash using IPFS daemon")
+        val gateway = Random.shuffle(knownGateways).head
```
propensive: I think that determinism would be better, though I also get the idea that building a tool which builds in a third-party service without a SLA isn't a good idea. But we will make it configurable.

src/core/external.scala
```diff
@@ -40,7 +45,15 @@ object Ipfs {
...
+        log.warn(s"Trying $gateway")
```
propensive: Use `info` instead of `warn`.

src/core/external.scala
```diff
@@ -63,6 +76,15 @@ object Ipfs {
       _    =  TarGz.untar(in, path)
     } yield path.childPaths.head
 
+    private def getFileFromGateway(hash: Multihash, path: Path)(gateway: String)(implicit log: Log): Try[Path] = {
+      for {
+        _    <- path.mkParents()
+        data <- Http.get(Https(Path(gateway) / "api" / "v0" / s"get?arg=${hash.toBase58}&archive=true"), Map(), Set.empty)
```
propensive: Is this URL better than `https://<gateway>/ipfs/<hash>`? It avoids directories being displayed through an HTML web interface, right?

propensive (changes requested):

src/core/external.scala
odisseus: I don't want all the users to hit the same gateway all the time, so I think it is better to choose one at random. On the other hand, this feature is going to be used very rarely (I hope), and it should not make any significant impact on any gateway.

src/core/external.scala
odisseus: This state is an error, although a recoverable one. Therefore, I think the warn level is appropriate here.

src/core/external.scala
odisseus: This is [part](https://docs.ipfs.io/reference/api/http/#api-v0-get) of the standard HTTP interface of an IPFS node. This particular endpoint allows to download a directory as a single TAR file.

src/core/external.scala
odisseus: Does the configuration format support lists of values?

odisseus: See #1115.

src/core/external.scala
odisseus: The shuffling has been removed.

src/core/external.scala
odisseus: Done

src/core/external.scala
propensive: Yes, it does. We have an OGDL [de]serializer for `List[T]`.

src/core/external.scala
```diff
@@ -56,13 +72,24 @@ object Ipfs {
...
+    private[this] def getFile(hash: Multihash, path: Path): Try[Path] = for {
       _    <- path.mkParents()
       data =  api.get(hash)
       in   =  new ByteArrayInputStream(data)
       _    =  TarGz.untar(in, path)
     } yield path.childPaths.head
 
+    private[this] def getFileFromGateway(hash: Multihash, path: Path)(gateway: String)(implicit log: Log): Try[Path] = {
+      log.info(s"Accessing $gateway to retrieve $hash")
+      val query = Query & "arg" -> hash.toBase58 & "archive" -> "true"
```
propensive: Whilst the syntax for `Query` is novel, I think I would prefer something more similar to a `ListMap`, i.e. an ordered map. My main concern is the introduction of `&` as an operator. It needs to have weaker precedence than `->` in order to work, and this isn't so clear. I think it would make sense to write this as, `Query("arg" -> hash.toBase58, "archive" -> "true")` or `Query(ListMap("arg" -> hash.toBase58, "archive" -> "true"))`.

src/core/external.scala
```diff
@@ -56,13 +72,24 @@ object Ipfs {
       )
     }
 
-    private def getFile(hash: Multihash, path: Path): Try[Path] = for {
+    private[this] def getFile(hash: Multihash, path: Path): Try[Path] = for {
```
propensive, line 75: There needs to be a timeout on here, otherwise it will never terminate and will never fall back to using HTTP.

propensive (changes requested):
I think we need something extra in the logic for when to use HTTP and when to use IPFS:
- if the user has IPFS installed on the `PATH` or a Fury-managed installation, use it, falling back to HTTP if the request times out
- if not, and the user is doing an IPFS `add` operation, install IPFS into order to perform the operation
- if not, use the HTTP gateway without installing IPFS.

In other words: use IPFS if we have it, install it only if we need it, otherwise use HTTP, and use HTTP if IPFS times out.

src/core/external.scala
odisseus: That's a good point, but I'd rather do it in another pull request. In fact, I think these requests do terminate eventually, but this happens after 10 or so minutes due to gateway timeout on the server.

odisseus:
Regarding the usage of IPFS gateways vs setting up a local node, I think the local node should be used as much as possible. This will avoid the strain on the gateways and make the user independent of them. There is even an [opinion](https://medium.com/pinata/the-ipfs-gateway-problem-64bbe7eb8170) that public gateways do a disservice to the IPFS community.

src/core/external.scala
odisseus: The current syntax allows adding more parentheses to make the priority of operations more clear:
```
val query1 = Query & ("arg" -> hash.toBase58) & ("archive" -> "true")
val query2 = Query & ("arg", hash.toBase58) & ("archive", "true")
```

src/core/external.scala
odisseus: See #1116.

propensive:
> Regarding the usage of IPFS gateways vs setting up a local node, I think the local node should be used as much as possible. This will avoid the strain on the gateways and make the user independent of them. There is even an [opinion](https://medium.com/pinata/the-ipfs-gateway-problem-64bbe7eb8170) that public gateways do a disservice to the IPFS community.

Yes, I would rather use the node if we can. But if it's a choice between downloading a 10K file over HTTP and a large download of IPFS, which might not work due to network restrictions, then I would rather download with HTTP.

PR #1113, opened 2020-03-02
Better support for URI queries

odisseus:
This may affect #1110.


propensive:
No, The reason is not working is that the server doesn't currently handle it...

PR #1121, opened 2020-03-04
Fix GitHub authentication

propensive:


odisseus (changes requested):
Doesn't work for me.

In the browser window that opens automatically, I get just a JSON with a single "token" field. It is not clear how to proceed.

When I click on the link in the terminal, I get the following message:
```
Error: Server Error
The server encountered an error and could not complete your request.

Please try again in 30 seconds.
```

propensive:
Maybe it's because the server wasn't working at the time. (Reasonably likely.)

PR #1139, opened 2020-03-10
Set up a 10 seconds timeout for HTTP requests to the IPFS daemon

### LINKED ISSUE #1116: Set a timeout for IPFS operations
The operations that involve downloading data from IPFS (including the situations when public gateways are used) should have a reasonable time limit, around 1-2 minutes. This affects the situations when the local IPFS node cannot maintain good connections to the rest of the network, or a requested hash is not present in the IPFS.

odisseus:
Fixes #1116.
```
$ fury layer clone -l fury://QmRTBGgmjcmhvaGTh3v7vXdQSCWJgC88CRRg7yCYaMY2v7
 10.063 Could not resolve the hash using IPFS daemon. Cause: java.lang.RuntimeException: timeout (10000 ms) has been exceeded
 10.067 Accessing ipfs.io to retrieve QmRTBGgmjcmhvaGTh3v7vXdQSCWJgC88CRRg7yCYaMY2v7
 20.350 Could not resolve the hash using ipfs.io. Cause: java.net.SocketTimeoutException: Read timed out
 20.354 Accessing gateway.pinata.cloud to retrieve QmRTBGgmjcmhvaGTh3v7vXdQSCWJgC88CRRg7yCYaMY2v7
 30.592 Could not resolve the hash using gateway.pinata.cloud. Cause: java.net.SocketTimeoutException: Read timed out
 30.610 An IPFS operation timed out.
```

### DIFF
--- a/src/core/external.scala
+++ b/src/core/external.scala
@@ -107,7 +107,7 @@ object Ipfs {
   def daemon(env: Environment, quiet: Boolean)(implicit log: Log): Try[IpfsApi] = {
     log.note("Checking for IPFS daemon")
 
-    def getHandle(): Try[IPFS] = Try(new IPFS("localhost", 5001))
+    def getHandle(): Try[IPFS] = Try(new IPFS("localhost", 5001).timeout(10000))
 
     def init(ipfs: Path): Try[Unit] = Xdg.ipfsRepo.ifExists().map(Success(()).waive).getOrElse {
       sh"${ipfs.value} init".exec[Try[String]]().map(_ => ())
--- a/src/core/external.scala
+++ b/src/core/external.scala
@@ -64,6 +64,9 @@ object Ipfs {
             }
             result
         }
+      }.recoverWith {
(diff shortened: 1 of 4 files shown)


src/core/external.scala
```diff
@@ -64,6 +64,9 @@ object Ipfs {
             }
             result
         }
+      }.recoverWith {
+        case e: RuntimeException if e.getMessage matches "timeout \\(.+\\) has been exceeded" => Failure(IpfsTimeout())
```
propensive, line 68: It's horrible that we have to do this...

propensive (approved):

src/core/external.scala
odisseus: Java programmers seem to be reluctant to create new types of exceptions. I think they see it as unnecessary because the good practice says that application logic should not be based on exceptions.

propensive:
I just merged this into `issue/1069` and it's still failing in CI with a five-minute timeout after starting IPFS. Any chance you could take a look, @odisseus?

odisseus:
You have merged the changes into the code that is being built, but have you updated the bootstrapping version that builds the new code?

PR #1140, opened 2020-03-10
Package just the Fury global log

odisseus:
Various services on the Mac OS runner generate a lot of log files, which we never actually read.


propensive:
Is it really failing after *44m*?

odisseus:
Looks like some problem with the Github servers.

Issue #1135, opened 2020-03-09
UnknownCommand exception is not reported


odisseus:
Command line
```
$ fury clone -l propensive/gastronomy -d gastronomy-build
$ echo $?
1

```

Global log
```
2020-03-09.10:16:12.572 [fail] 11dfc The command 'clone' was not recognized.
```

Issue #1088, opened 2020-02-20
Building a nonexistent module with a slash in the name triggers build of the selected module


odisseus:
```
$ fury -m foobar
The module has not been specified.

$ fury -m foo-bar
The module has not been specified.

$ fury -m foo/bar --output linear
Starting compilation of module fury/io
Successfully compiled module fury/io
...
Starting compilation of module fury/frontend
Successfully compiled module fury/frontend
Starting compilation of module fury/project
Successfully compiled module fury/project

$ fury -m / --output linear
Starting compilation of module fury/io
Successfully compiled module fury/io
...
Starting compilation of module fury/frontend
Successfully compiled module fury/frontend

```

szymon-rd:
The same issue appears for building projects. Will try to fix it as well.

szymon-rd:
Actually, PR was closed not merged, reopening issue due to my mistake.

odisseus:
The error message does not appear if the wrong module name contains a slash, like `foo/bar`.

odisseus:
This issue returned after PR #1223 was merged. Every kind of problem with the module name argument is now treated as if that argument were missing, and the default module is substituted.

odisseus:
PR #1282 has [brought](https://github.com/propensive/fury/pull/1282/files#diff-585434c8ce4e375972307bf1bc53ab50L201) this issue back again. I see a pattern here...

Issue #776, opened 2019-12-12
permissions that were obviated persist in the policy file


odisseus:
## Steps to reproduce

1. Run any application module that requires some permissions and make sure it runs correctly.
2. Obviate any of the required permissions.
3. Run the application module again.

**Expected behaviour:** a security exception is thrown
**Actual behaviour:** the application module runs just like the first time.

## Workaround
1. Make sure Fury is not running
2. Remove all Java policy files (usually stored in `~/.cache/fury/policies`)
3. Remove the corresponding entries from the global policy file (usually `~/.config/fury/policy.fury`).

propensive:
The actual behavior is correct, but there needs to be a flag to additionally revoke a privilege at the same time as it's obviated. And there should be a separate `revoke` command.

PR #1147, opened 2020-03-13
Added an integration test for various error reports.

odisseus:
This integration test is related to issues #1088, #1135 and #1148. It may be affected by #776.


propensive:
Do you know why this is passing on `macos-latest` when the `check` file contains pseudo-output?

odisseus:
This is a pending test, so its failure doesn't affect the outcome of the build.

PR #1156, opened 2020-03-15
Don't repackage directory entries from dependency JARs into a fat JAR

odisseus:
This should make the duplicate entries in a resulting JAR file much less frequent.

Nevertheless, the problem isn't completely solved, as the dependency JARs might contain _files_ with duplicate names.

### DIFF
--- a/test/passing/fat-jar/check
+++ b/test/passing/fat-jar/check
@@ -17,7 +17,7 @@ Starting compilation of module bar/app
 Successfully compiled module bar/app
 Hello World
 Saving JAR file out/bar-app.jar using UTF-8
-Wrote 18.9 MiB to out/bar-app.jar
+Wrote 18.8 MiB to out/bar-app.jar
 0
 bar-app.jar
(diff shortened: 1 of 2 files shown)


propensive (approved):

test/passing/fat-jar/check
```diff
@@ -17,7 +17,7 @@ Starting compilation of module bar/app
 Successfully compiled module bar/app
 Hello World
 Saving JAR file out/bar-app.jar using UTF-8
-Wrote 18.9 MiB to out/bar-app.jar
+Wrote 18.8 MiB to out/bar-app.jar
```
propensive, line 20: Is this reduction in size because we're not including duplicate directory entries in the JAR file?

odisseus: So it seems. The size of this JAR [used to be](https://github.com/propensive/fury/pull/1144/files#diff-1953202c152089333eccbeb31459c44dL20) 18.8 MiB before  PR #1144.

PR #1159, opened 2020-03-16
Fixed the argument to define permission requirements without granting 

odisseus:


test/passing/permission-granting/check
```diff
@@ -8,6 +8,18 @@ Set current module to app
...
+│ Hash │ Class                        │ Target                     │ Action │
+├──────┼──────────────────────────────┼────────────────────────────┼────────┤
+│ 75e  │ java.util.PropertyPermission │ scala.copyright.string     │ read   │
+│ 0bc  │ java.util.PropertyPermission │ scala.maven.version.number │ read   │
+│ d42  │ java.util.PropertyPermission │ scala.time                 │ read   │
+│ 4d1  │ java.util.PropertyPermission │ scala.version.number       │ read   │
+└──────┴──────────────────────────────┴────────────────────────────┴────────┘
+You can grant these permissions with,
+
+  fury permission grant -P <permission hash>
```
propensive: This looks fragile, but I can't think of a better way right now.

test/passing/permission-granting/script
```diff
@@ -1,19 +1,22 @@
+rm ~/.config/fury/policy.conf
+rm ~/.cache/fury/policies/*
```
propensive: Deleting the policies shouldn't be necessary for anything.

test/passing/permission-granting/script
```diff
@@ -1,19 +1,22 @@
...
 fury project add -n hello-world
 fury module add -n app -c scala/compiler -t application -M HelloWorld
+fury source add -d src
+mkdir -p src
+echo 'object HelloWorld extends App { println("All permissions in order!\n") }' > src/hw.scala
 fury permission require -0 -C java.util.PropertyPermission -T scala.maven.version.number -A read
 fury permission require --no-grant -C java.util.PropertyPermission -T scala.version.number -A read
 fury permission require -C java.util.PropertyPermission -T scala.time -A read -0
 fury permission require -C java.util.PropertyPermission -T scala.copyright.string -A read --no-grant
+fury build run --output linear | sed "s/\x1B\[\([0-9]\{1,2\}\(;[0-9]\{1,2\}\)\?\)\?[mGK]//g"
```
propensive: What is this removing? If it's ANSI control characters, wouldn't it be better just to set the config theme first?

odisseus: I copied the regular expression from [Stack Overflow](https://unix.stackexchange.com/a/111936). It is necessary because some ANSI codes are hardcoded in the library that prepares tabular output.

test/passing/permission-granting/script
odisseus: Ok

test/passing/permission-granting/script
propensive: Oh, that makes sense, though it's a bit annoying. I'll open an issue on Escritoire.

test/passing/permission-granting/check
odisseus: I have removed the table from the check file and simplified the regex.

PR #1173, opened 2020-03-18
Cleanup of `Layer` to use `Qm` codes everywhere, lazily

propensive:
This is still work-in-progress. Some features are still not fully
re-implemented, but it compiles, and some thins work.


src/core/layer.scala
```diff
@@ -35,265 +37,168 @@ case class Layer(version: Int,
...
+    val version = Try(ogdl.version().toInt).getOrElse(1)
+    if(version < CurrentVersion) {
+      log.note(msg"Migrating layer file from version $version to ${version + 1}")
+      migrate((version match {
+        case 0 | 1 | 2 =>
+          log.fail(msg"Cannot migrate from layers earlier than version 3")
+          // FIXME: Handle this better
+          throw new Exception()
+        case 3 =>
+          ogdl.set(schemas = ogdl.schemas.map { schema =>
```
odisseus: Does migration from version 3 still work? If not, this seems like a good opportunity to drop it. I don't think there are any layers that still use version 3.

propensive: I think it dosen't work. I'm going to upgrade to version `6` too, which will eliminate the concept of schemas completely. I could probably write a migration from `5` -> `6`, but more likely I'll just take the opportunity to start afresh, and I'll take the weekend to redefine everything.

src/core/install.scala
```diff
@@ -51,6 +53,27 @@ object Install {
...
+  private def iconDirs: List[Path] = List(16, 48, 128).map { s =>
+    Path("icons") / "hicolor" / str"${s}x${s}" / "apps" / "fury-icon.png"
+  }
+
+  private def desktopInstall(env: Environment)(implicit log: Log): Try[Unit] = for {
+    _       <- ~log.info(msg"Installing handler for fury:// URLs")
+    file    <- ~(Xdg.dataHome / "applications" / "fury.desktop")
+    _       <- ~file.writeSync(desktopEntry)
+    _       <- iconDirs.traverse { p => p.in(Installation.etcDir).copyTo(p.in(Xdg.dataHome)) }
+    _       <- sh"xdg-mime default fury.desktop x-scheme-handler/fury".exec[Try[String]]
```
odisseus: This probably won't work on Mac OS.

Makefile
```diff
@@ -83,10 +83,19 @@ tmp/bin/ng.py:
 	 chmod +x "$@" && \
 	 printf "done\n" || printf "failed\n"
 
-dist/fury.tar.gz: tmp/.version tmp/lib/fury.jar tmp/bin/fury tmp/bin/ng.c tmp/bin/ng.py tmp/bin/procname.c tmp/script/_fury
+tmp/etc: doc/logo/render_1000px.png
+	@printf "Resizing logo for Fury icons..." && \
+	 mkdir -p tmp/etc/icons/hicolor/128x128/apps tmp/etc/icons/hicolor/48x48/apps tmp/etc/icons/hicolor/16x16/apps && \
+	 convert doc/logo/render_1000px.png -resize 16x16 tmp/etc/icons/hicolor/16x16/apps/fury-icon.png && \
```
odisseus: This probably won't work if the user doesn't have ImageMagick installed.

src/core/install.scala
propensive: I think the XDG layout and variables are standard on Mac OS, though it would be useful to have someone else check this...

Makefile
propensive: True. I've committed the converted icons to the repository, and updated the makefile to have a separate `icons` target which will build them, if called explicitly, whilst not requiring this to be part of the standard installation.

propensive:
Aggressively merging this before the start of the week, even though CI isn't working. I'll try to fix that as soon as possible.

PR #1184, opened 2020-03-23
Untangle BSP target ID from BSP client

odisseus:


propensive (commented):
How "expected" are these thrown exceptions? Roughly what would have to happen for Fury to get into the illegal state?

Generally we're not _throwing_ exceptions, but I suppose the BSP API doesn't allow us to do anything better. Can we rely on the consumer of the BSP API to handle those exceptions?

propensive:
Are we sure that the community build is actually building anything successfully?

odisseus:
I'm going to remove those exceptions before merging this pull request. The code that handles notifications should not throw under normal conditions.

propensive (approved):

Issue #1197, opened 2020-03-31
The fat JAR includes META-INF contents from dependency JARs


odisseus:
In particular, the JAR build for Fury contains manifest signatures from some other JAR. This [causes](https://stackoverflow.com/questions/999489/invalid-signature-file-when-attempting-to-run-a-jar#6743609) errors during installation.

propensive:
I made a change recently to perform classpath-equivalent "shadowing" of identically-named files when constructing the fat JAR, as I was getting some "duplicate entry" errors when constructing the JAR. I didn't change the rule to skip all `META-INF` entries, though, so I'm not sure why this has come from.

#1198 looks like it is intended to just accept these files will be present, rather than remove them. Is that right?

odisseus:
PR #1198 is just a workaround to make the current bootstrapping version usable.

propensive:
I think this problem went away, didn't it?

odisseus:
So it seems, but I think this issue needs further investigation. It looks like there are some conditions, when Fury would package leftovers of other builds into the fat JAR.

propensive:
Yes, I think there's still something strange going on there.

propensive:
Having said that, I'm happy that the main problem is fixed, and we shouldn't see zipfile errors again, so I'm going to close this.

PR #1198, opened 2020-03-31
Preload layers before building

odisseus:
This pull request is related to the issue #1197.

### DIFF
--- a/.fury.conf
+++ b/.fury.conf
@@ -7,7 +7,7 @@
 #
 # For more information, please visit https://propensive.com/fury/
 #
-layerRef	QmSajASDC5iFvRRMJU76pGM8dFnwKzSGuyU3PSzuE4RjUz
+layerRef	QmVjt5jkxX2LLbr5MnuVdskwiwskRkfGTnPyZ163k3AuDr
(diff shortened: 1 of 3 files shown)


propensive (changes requested):
I think those `zip` commands will be slow. Each one will rewrite the whole 45MB file.

The problem, I think, is that for some reason, these files are not being filtered out. I'm not sure why because there's a condition which checks whether `META-INF` occurs in the entry name which should forbid it from being included in the fat JAR.

odisseus:
I don't think these commands are going to be slow. Each one takes less than a second on my computer. For comparison, unzipping the entire archive takes about 10 seconds.

odisseus:
I have modified the layer definition so thar the repo at `gh:propensive/fury` points to the master branch, commit d7e7d44. This fixed the compilation errors.

Also, the spurious files in META-INF seem to have gone, so I removed the workaround to delete them.

propensive:
Ah, that explains everything then. It was pointing to the wrong version...

propensive:
We even got 8 integration tests to pass! I'll do some work on fixing the rest.

propensive (approved):

PR #1201, opened 2020-03-31
tempoary hacky bash completion script

coconutpalm:
Here's a completion script for Bash that just reformats the zsh completion output.  It's hacky but it works.  :-)

### DIFF
--- a/etc/completion/bash/.fury.bash
+++ b/etc/completion/bash/.fury.bash
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+_fury_completions() {
+    # Temporary hack: reformat the zsh completions for bash
+    CHOICES=$(fury completion --shell zsh --paramNo $COMP_CWORD -- $COMP_LINE | \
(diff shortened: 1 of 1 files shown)


propensive:
Thank you! That looks like a great stop-gap until I can implement the new version. But why did you close it? ;)

coconutpalm:
Signing the committer agreement requires getting my company's legal department involved.  And lawyers...

But feel free to use the ideas and write your own.. :-)

coconutpalm:
That said, I love Fury's idea and think it could save my company's teams a lot of work.  Dependencies bite us all the time.  Maybe in the future...

propensive:
Ah, I see. :( Depending on how I get on with some of the ideas I have planned for Exoskeleton (which is looking quite hopeful as a full `bash`/`zsh`/`fish` completion system, but isn't ready yet), I may seek inspiration for a hack like yours. Sorry that the CLA made it awkward this time, though...

PR #1242, opened 2020-04-07
Fixes and tests for OGDL serialization of collections

### LINKED ISSUE #1202: Serialization of sorted sets is broken
This is why integration tests fail with permission errors.

Aparently, this problem was [caused](https://github.com/propensive/fury/commit/d7e7d44ff60c97b8c156d59ba811acbf21950c08#diff-21de65a622bc41008e02fc83d48d3e67R85) by commit d7e7d44ff60c97b8c156d59ba811acbf21950c08.

Underlying exception ("Element id not found"):
```
fury.ogdl.OgdlException
	at fury.ogdl.Ogdl.selectDynamic(ogdl.scala:44)
	at fury.ogdl.OgdlWriter$.$anonfun$treeSet$2(ogdlwriter.scala:86)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
	at scala.collection.Iterator.foreach(Iterator.scala:941)
	at scala.collection.Iterator.foreach$(Iterator.scala:941)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.map(TraversableLike.scala:238)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at fury.ogdl.OgdlWriter$.$anonfun$treeSet$1(ogdlwriter.scala:82)
	at fury.ogdl.Ogdl$.apply(ogdl.scala:68)
```

odisseus:
This pull request should fix #1202.

## The problem 
The rules of OGDL serialization were updated so that elements of lists and sets are stored as key-value pairs, where the key ("index") is one of the fields of the element object, and the value is a vector of all the other fields.
Example for `case class Foo (id: Id, data1: T1, data2: T2, ...)`:
```
collection-id  -->  ( * , * , * , ...)
                      |
                      \- Ogdl(id)  -->  ( * , * , ...)
                                          |   |
                                          \---+-- "data1" --> Ogdl(data1)
                                              |
                                              \-- "data2" --> Ogdl(data2)
```

However, this approach has two problems:
1. How to choose the index if the type `Foo` has no fields at all (e. g. `type Foo = String`)?
2. How to store the index if it is a complex type itself (e. g. for `case class Bar (id: Foo, data: T)`)?

## The solution

In the case when collection elements are primitive values (or strings), the value of each element is used as its index:
```
collection-id  -->  ( * , * , * , ...)
                      |   |   |
                      \---+---+-- element1.toString --> Ogdl(element1)
                          |   |
                          \---+-- element2.toString --> Ogdl(element2)
                              |
                              \-- element3.toString --> Ogdl(element3)
```

In the case when the index is an object which itself has fields, an artificial OGDL node is constructed, which stores both the index object and the rest of the collection element:

```
collection-id  -->  ( * , * , * , ...)
                      |
                      \-- "kvp" --> ( * , * )
                                      |   |
                                      \---+-- "id" --> Ogdl(id)
                                          |
                                          \-- "value" --> ( * , * , ...)
                                                            |   |
                                                            \---+-- "data1" --> Ogdl(data1)
                                                                |
                                                                \-- "data2" --> Ogdl(data2)
```
For the latter case, the node names "kvp" and "value" are hardcoded. These are special names for the nodes that must be created for the purpose od serialization, and don't correspond directly  to any Scala object.

See the tests for live examples. However, to run these tests you'll have to restore a test module which has been lost some time ago.

### DIFF
--- a/src/core/source.scala
+++ b/src/core/source.scala
@@ -29,6 +29,7 @@ object Source {
   implicit val ogdlReader: OgdlReader[Source] = src => unapply(src.id()).get // FIXME
   implicit val ogdlWriter: OgdlWriter[Source] = src => Ogdl(Vector("id" -> Ogdl(src.key)))
   implicit val parser: Parser[Source] = unapply(_)
+  implicit val index: Index[Source] = FieldIndex("id")
   implicit val keyName: KeyName[Source] = () => msg"source"
 
   implicit val sourceDiff: Diff[Source] =
--- a/src/model/ids.scala
+++ b/src/model/ids.scala
@@ -450,7 +450,7 @@ sealed abstract class ScopeId(val id: String) extends scala.Product with scala.S
 object Grant {
   implicit val ord: Ordering[Grant] = Ordering[String].on[Grant](_.permission.hash)
   implicit val stringShow: StringShow[Grant] = _.digest[Sha256].encoded
-  implicit val index: Index[Grant] = Index("permission")
+  implicit val index: Index[Grant] = FieldIndex("permission")
 }
 
 case class Grant(scope: Scope, permission: Permission)
--- a/src/ogdl-test/testreader.scala
+++ b/src/ogdl-test/testreader.scala
@@ -28,6 +28,9 @@ object OgdlReaderTest extends TestApp {
   private[this] case class Foo(bar: String, baz: Int)
   private[this] implicit val ord: Ordering[Foo] = Ordering[Int].on[Foo](_.baz)
 
+  private[this] case class Foo2(bar: Option[String], baz: Int)
+  private[this] implicit val ord2: Ordering[Foo2] = Ordering[Int].on[Foo2](_.baz)
+
   private[this] case class Quux(handle: String, data: SortedSet[String])
 
   override def tests(): Unit = {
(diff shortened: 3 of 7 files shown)


propensive:
Looks like good progress. I'll have a closer look tomorrow.

propensive:
I've just read the description, but were we actually serializing `SortedSet`s of primitives anywhere?

odisseus:
I'm not sure about sorted sets, but  e. g. [`Alias`](https://github.com/propensive/fury/blob/master/src/model/ids.scala#L539) contains a list of strings. The problem with serializing elements of primitive types is common for lists and sets.

propensive:
Yes, `Alias` is a good example, and maybe not the only one we will ever have.

Is there anything which stops us serializing `List("foo", "bar", "baz")` as,
```
foo     bar     baz
```
(where those gaps are tabs)?

That's about as efficiently as they can be stored.

odisseus:
I'm afraid it's impossible to create an OGDL [node](https://github.com/propensive/fury/blob/master/src/ogdl/ogdl.scala#L39) that would serialize into a sequence of primitive values. Each node must contain a list of nested nodes, ad it is not possible to store a value without attaching it to another node:
```
case class Ogdl(values: Vector[(String, Ogdl)])
```

propensive:
But why not,
```
List("foo", "bar", "baz") -> Ogdl(Vector(("foo", Ogdl(Vector(("bar", Ogdl(Vector(("baz", Ogdl(Vector()))))))))))
```
It's isomorphic to the internal structure of the `List`...

odisseus:
Done

propensive:
Thank you!

PR #1245, opened 2020-04-08
Restored the main class which is expected by the launcher script.

odisseus:


propensive (approved):
I'm confused about how this went missing, but if this fixes things, great.

PR #1248, opened 2020-04-09
Use the layer diff mechanism to check if the layer has been modified

### LINKED ISSUE #1247: Changes to tree sets in the layer object are not saved
A layer with a single module is considered equal to a layer with no modules at all .This breaks most integration tests.
Here's an example of a failing test for a tree set of projects (which is part of a layer object):

```
test("Unequal projects") {
  val x = Project(ProjectId("scala"),
    TreeSet(),
    None, LicenseId("unknown"), "", None)
  val y = Project(ProjectId("scala"),
    TreeSet(Module(ModuleId("compiler"), Compiler, None, None, List(), ModuleRef("java/compiler"), TreeSet(),
      TreeSet(), TreeSet(), TreeSet(), TreeSet(), None, TreeSet(), TreeSet(), TreeSet(), false, TreeSet(), false)
    ),
    Some(ModuleId("compiler")), LicenseId("unknown"), "", None)
  //x == y //false
  TreeSet(x) == TreeSet(y) //true!
}.assert(_ == false)
```

This behaviour turned out to be caused by the way tree sets are compared: https://github.com/scala/bug/issues/10741
Tree set elements are considered equal when comparing sets if those elements are equal for the purpose of ordering, and projects are equal for the purpose of ordering if their IDs are equal.

This didn't cause any harm before, because a layer cannot have two projects with the same ID.
However, when comparing two different layers, this causes unexpected results.

odisseus:
Fixes #1247.

### DIFF
--- a/src/core/layer.scala
+++ b/src/core/layer.scala
@@ -175,8 +175,7 @@ object Layer extends Lens.Partial[Layer] {
     base     <- get(baseRef, conf.published)
     layer    <- ~base.copy(previous = Some(conf.layerRef))
     previous <- retrieve(conf)
-
-    _        <- if(previous.copy(previous = None) != layer.copy(previous = None))
+    _        <- if(!Layer.diff(previous.copy(previous = None), layer.copy(previous = None)).isEmpty)
                     store(layer).flatMap { baseRef => saveFuryConf(conf.copy(layerRef = baseRef), layout) }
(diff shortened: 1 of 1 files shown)


propensive (approved):
This might not be the fastest way to check, but it will work.

Issue #1222, opened 2020-04-05
tab completion on `dependency add` causes error


propensive:
When typing,
```
fury dependency add -m<tab>
```
an error is reported.

The error is fine, but it shouldn't be printed at all, because the system should be in "completion" mode, which should mean that no output other than completions should be produced. I'm not sure why it is.

odisseus:
```
_arguments:comparguments:325: invalid argument: The argument '' was not a valid value for the parameter '-m/--module'.
```

odisseus:
Looks like this issue is caused by a bug somewhere else, probably in the code that handles command completions.

## Stack trace
```
An unexpected error occurred:
exoskeleton.InvalidArgValue: the value '-m/--module' is not valid for the parameter --
    at exoskeleton.ParamMap.get(parameters.scala:92)
    at fury.core.Cli.preview(cli.scala:184)
    at fury.DependencyCli.$anonfun$add$10(dependency.scala:102)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$8(dependency.scala:101)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$6(dependency.scala:100)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$4(dependency.scala:99)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$3(dependency.scala:98)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$2(dependency.scala:97)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.$anonfun$add$1(dependency.scala:96)
    at scala.util.Success.flatMap(Try.scala:251)
    at fury.DependencyCli.add(dependency.scala:95)
    at fury.FuryMenu$.$anonfun$menu$29(menu.scala:73)
    at fury.core.Menu.apply(structure.scala:82)
    at fury.core.Menu.apply(structure.scala:80)
    at fury.core.Cli$.asCompletion(cli.scala:101)
    at fury.FuryMenu$.$anonfun$menu$26(menu.scala:67)
    at fury.core.Menu.apply(structure.scala:82)
    at fury.FuryServer$.invoke(main.scala:49)
    at fury.FuryServer$.$anonfun$run$2(main.scala:92)
[...]

PR #1265, opened 2020-04-14
Refactored the commands related to dependencies

odisseus:
The commands related to dependency management are refactored to keep both command completions and meaningful error messages.

The commit 1ba336e47a26782370f75d4625f54f75e250ad52 is reverted, because caused the issue #1088 to reappear.

This pull request is related to issues #1088 and #1222.

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -101,6 +101,7 @@ case class ConfigCli(cli: Cli)(implicit log: Log) {
     _      <- ~log.infoWhen(!raw)(conf.focus())
     _      <- ~log.rawln(table)
   } yield log.await()
+
 }
 
 case class AboutCli(cli: Cli)(implicit log: Log) {
--- a/src/core/cli.scala
+++ b/src/core/cli.scala
@@ -180,10 +180,11 @@ class Cli(val stdout: java.io.PrintWriter,
   def action(blk: Call => Try[ExitStatus])(implicit log: Log): Try[ExitStatus] = call().flatMap(blk)
 
   def peek(param: CliParam): Option[param.Type] = args.get(param.param).toOption
-  
   def preview(param: CliParam)(default: Option[param.Type] = None): Try[param.Type] = {
     val result = args.get(param.param)
-    if(default.isDefined) result.recover { case _: Exception => default.get } else result
+    if(default.isDefined) result.recover {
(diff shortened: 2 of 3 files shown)


propensive (commented):
I don't remember the details of the original change, though there were certain tab-completions that would fail, printing unwanted error output to STDOUT, which could not be interpreted by the completions system.

odisseus:
I have found a way to solve those problems. I have updated just a couple commands so far, but I'll refactor all of them if my solution works well.

propensive (changes requested):
Thanks for working on this. I think we need to try to reduce the amount of code we have here. I think that certainly some of the logic could be factored into methods on one of our model objects. I don't want to mix up front-end code (e.g. things like `cli` and `ModuleArg`) with backend code, but we could have a method on `Layer` which gets the current module/project. Something like this:
```
def moduleRef(project: Option[ProjectId] = None, module: Option[ModuleId] = None): Try[ModuleRef] 
```
This won't be the whole story, of course, but I think there's a lot of improvement to be made here.

odisseus:
The actual problem lies not in the repetitive code, but in the fact that the same code must handle bad arguments in two different ways. 

When the command completion is constructed, we want to hide any exceptions and substitute failures with nil values (usually `None` or empty list) so that the completion hints can be constructed further. 

When the command is actually executed, we want to interrupt execution on the first failure, and produce a descriptive message about the nature of the failure.

As a consequence, each potential failure to resolve a value against the context built by the previous steps (e. g. project depending on the layer, module depending on the project) must be both substituted and stored. This is the reason why the code that resolves e. g. the project cannot be inlined into the main for-comprehension, even though it returns a `Try`.

I agree that the code which handles the commands is quite repetitive, and it must be possible to extract common idioms (such as getting a valid `ModuleRef`). However, cleaning up the CLI is a huge chunk of work that lies outside the scope of this pull request. My goal here is just resolving the contradiction between #1088 and #1222. Besides, repetitive as it is, I believe my code is slightly better and clearer than in the master.

propensive:
Yes, I agree there. I've got quite a lot of ideas for fixing that code which will have to wait until after the next release. So I can either merge this now, or wait for you to replicate the fixes elsewhere. Please go ahead and merge when ready!

PR #1286, opened 2020-04-21
Restored the main class

odisseus:


propensive (approved):
I think it's literally character-for-character the same as my version, but you got there seven minutes earlier. ;)

PR #1301, opened 2020-04-27
New version of Probably

odisseus:


propensive (changes requested):
If you use the most recent version of Probably, check out the style of tests in `src/test/tests.scala`. I have furthermore added nice hierarchical support for "suites", which I'm just about to document. (There are no examples of its usage, yet.)

odisseus:
In order to make the tests compile, I had to perform the following operations on the layers:
* updated **platform** from version 1.5 to 1.7 (not sure if that was absolutely necessary);
* removed the module **macros** from **probably**;
* added dependencies on **magnolia** and **gastronomy** to **probably**;
* updated the reference to [probably](https://github.com/propensive/probably) repository to the master branch and the latest commit.

PR #1305, opened 2020-04-28
Fetch all

propensive:


odisseus:
This pull request probably does resolve the current problems with fetching, but it also introduces a potential performance issue. In a very large repository, there might be hundreds of active branched, and still more stale ones. It is pointless to fetch all of them when only one is actually needed.

propensive:
I think this is probably why it changed in the first place. Let's fix that later, though.

PR #1306, opened 2020-04-28
Write the updated layer after the state of repos has changed

odisseus:


propensive (approved):
Really? *That* was the bug...? :facepalm:

odisseus:
I'm afraid that's not the only bug related to updating the repo state. Fetching custom branches from a remote repository is probably still broken.

PR #1310, opened 2020-04-29
Rewrite of `checkin` and `checkout`

propensive:


src/core/git.scala
```diff
@@ -85,8 +76,29 @@ case class GitDir(env: Environment, dir: Path) {
...
 
+  def writePrePushHook(): Try[Unit] = for {
+    file <- Try((if((dir / ".git").exists()) dir / ".git" else dir) / "hooks" / "pre-push")
+    _    <- file.writeSync(
+              """|#!/bin/sh
+                |remote="$1"
+                |url="$2"
+                |while read local_ref local_sha remote_ref remote_sha
+                |do
+                |  git --no-pager diff --name-only $local_ref..$remote_ref -- .fury.conf && fury layer share
```
odisseus: I'm not sure it is a good idea to add Git hooks without asking the user. Automatically sharing all the intermediate layers doesn't seem like a good idea either.

propensive: I would argue that as *we* are checking out the repository, it's more acceptable, though I agree that we should log this fact.

I would also like to *ask* the user before running the `share` command. I'll make these changes before merging.

odisseus: It is acceptable to add any hooks to a repository which is completely managed by Fury (that is, created as a side effect of `fury repo add`). It is a bad idea to add or overwrite any hooks in any other repositories (including the local repository in the layout's base directory). 

A forked repository is sort of a middle ground, because it is controlled by Fury at its creation, but the user will probably modify its state directly, perhaps even adding additional hooks.Nevertheless, I can agree with Fury creating its own hooks when the repository is being forked.

By the way, there is another possible issue: the share operation might just fail for a number of reasons. In my opinion, this shouldn't abort the push.

odisseus:
I tried to execute the checkout command and got this:
```
% fury repo checkout -r fury
An error occurred while running: git clone gh:propensive/fury --branch=issue/1185 /home/virtuslab/devel/fury/.tmp

Cloning into '/home/virtuslab/devel/fury/.tmp'...
ssh: Could not resolve hostname gh: Name or service not known
fatal: Could not read from remote repository.
```

propensive:
This is confusing me a bit. It's doing exactly what the code says it should do, and we're storing the "simplified" repository name in the layer file. However, I can't work out why this isn't a problem in lots of other places, and why this works for me. I will, in any case, fix it.

src/core/git.scala
propensive: I think you should categorize a `fury repo check`ed`out` repository as essentially the same a fork. It's _partially_ managed by Fury. Though a user may still have a Fury build in a repository that's checked out with `git`, and in that case I will probably add an explicit command to add the pre-push hook.

odisseus: A repository should be just a repository of source files, be it remote, checked out or forked. Even if a repository happens to contain a `.fury.conf` file pointing to a layer, it should be treated as just a file. If the developer wanted to use that layer as a layer, he would have imported it into the current layer.

odisseus:
Eliminating the `--https` flag is a great idea, and I think those changes can be cherry-picked and merged separately from this pull request.

src/core/git.scala
```diff
@@ -35,12 +35,21 @@ import java.util.zip.ZipInputStream
 
 object GitDir {
 
+  private var useHttps: Boolean = false
```
odisseus, line 38: I think it's a bad idea to force HTTPS for all remotes, especially without a way to restore the previous state. Each particular remote might be accessible in just one way. Consider the situation when one remote is a LAN server without a HTTPS interface, and another remote is on Github, but outward SSH connections are blocked by the LAN firewall.

Why not just try both ways for each remote?

PR #1317, opened 2020-05-01
Fix reading older layers

odisseus:
Copy-pasted from PR #1310.

### DIFF
--- a/src/core/layer.scala
+++ b/src/core/layer.scala
@@ -163,7 +163,7 @@ object Layer extends Lens.Partial[Layer] {
   private def lookup(ref: IpfsRef): Option[Layer] = cache.synchronized(cache.get(ref))
   implicit val stringShow: StringShow[Layer] = store(_)(Log.log(Pid(0))).toOption.fold("???")(_.key)
 
-  val CurrentVersion: Int = 7
+  val CurrentVersion: Int = 8
 
   def set[T](newValue: T)(layer: Layer, lens: Lens[Layer, T, T]): Layer = lens(layer) = newValue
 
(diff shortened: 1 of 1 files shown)


src/core/layer.scala
```diff
@@ -322,10 +322,14 @@ object Layer extends Lens.Partial[Layer] {
...
       migrate((version match {
+        case 7 =>
+          Try(ogdl.set(repos = ogdl.repos.map { repo =>
+            repo.set(remote = repo.repo)
+          })).getOrElse(ogdl)
         case 6 =>
           // FIXME: This is a lazy solution
           Try(ogdl.set(repos = ogdl.repos.map { repo =>
-            repo.set(branch = repo.track)
+            repo.set(branch = repo.track, remote = repo.repo)
```
propensive: This line change isn't required. Each update should be applied incrementally in turn.

src/core/layer.scala
```diff
@@ -322,10 +322,14 @@ object Layer extends Lens.Partial[Layer] {
     if(version < CurrentVersion) {
       log.note(msg"Migrating layer file from version $version to ${version + 1}")
       migrate((version match {
+        case 7 =>
```
propensive, line 325: This should be close to identical to the changes in #1310.

propensive (approved):

src/core/layer.scala
odisseus: Yes, that's where i took these changes.

src/core/layer.scala
odisseus: Reverted

PR #1324, opened 2020-05-03
Include all `scala.*` properties in standard privileges

propensive:


odisseus:
Looks good, but the integration tests for permission granting will probably break.

PR #1329, opened 2020-05-03
Fix save-jar and fat-jar tests

propensive:


.fury.conf
```diff
@@ -1,13 +1,13 @@
...
-#  
+#
 # To start using Fury with this project, install Fury and run,
 #
 #   fury layer init
 #
 # For more information, please visit https://propensive.com/fury/
 #
-layerRef	QmRi4LSdm2pM9CUtZiU4Tw2ugfytUpRm81jcqAuEAdRA4J
+layerRef	QmRPnVfchh6LiAiRjZ18v5tXrS1WQvPR6bCMvrtKYaV6sa
```
odisseus, line 10: The new [layer](https://gateway.pinata.cloud/ipfs/QmRPnVfchh6LiAiRjZ18v5tXrS1WQvPR6bCMvrtKYaV6sa) does not contain any references to this repository.  Moreover, it contains references to local source directories.

It will work correctly only if the focus file is located inside a local copy of Fury's sources.

PR #1334, opened 2020-05-04
Don't output logs which differ every time in `layer-traversal`

propensive:


odisseus:
I don't  think this is a good solution.

It is very important to check that creating a layer with the same contents always yields the same hash. If this rule is broken, the deduplication property of IPFS will be lost, and both local caches and Pinata storage will be flooded by thousands of equivalent layers.

PR #1335, opened 2020-05-04
Tests fix

odisseus:
I have updated two of the three integration tests that still fail. The third one fails because of an regression that needs to be fixed in the code (see comments to #1088).

### DIFF
--- a/test/passing/layer-traversal/check
+++ b/test/passing/layer-traversal/check
@@ -4,14 +4,8 @@ Set current module to compiler
 Set current project to library
 Setting default compiler for project library to scala/compiler
 Set current module to core
-Cloning repository at ../library
-Checking out src/core from repository library
-Starting the BSP launcher...
-Opening a BSP server connection
-BSP server is listening for incoming connections
-Starting compilation of module library/core
-Successfully compiled module library/core
 0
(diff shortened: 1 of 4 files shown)


test/passing/layer-traversal/check
```diff
@@ -4,14 +4,8 @@ Set current module to compiler
...
 Set current module to core
-Cloning repository at ../library
-Checking out src/core from repository library
-Starting the BSP launcher...
-Opening a BSP server connection
-BSP server is listening for incoming connections
-Starting compilation of module library/core
-Successfully compiled module library/core
 0
+IPFS ref: fury://QmTbRCxBkeAF9NzS24LyoC7iDsSFsuB46EHRiWHUZyiG16
```
propensive, line 8: So we can actually get this to be stable? I couldn't see what else changed to ensure this... but if it's fine, we can merge it.

propensive (approved):

test/passing/layer-traversal/check
odisseus: I removed the references to Git repositories. All other elements (projects, imported layers etc.) are always the same if they are created using the same set of commands.

Issue #1330, opened 2020-05-03
regression on `bad-commands` test


propensive:
This appears to have stopped working again.

odisseus:
This is a duplicate of #1088.

odisseus:
The regression seems to be caused by this change: https://github.com/propensive/fury/pull/1282/files#diff-585434c8ce4e375972307bf1bc53ab50L201

PR #1337, opened 2020-05-04
Restored the Cli.preview method.

odisseus:
This pull request is related to the recent regression on issue #1330.

### DIFF
--- a/src/core/cli.scala
+++ b/src/core/cli.scala
@@ -220,9 +220,13 @@ class Cli(val stdout: java.io.PrintWriter,
   def action(blk: Call => Try[ExitStatus])(implicit log: Log): Try[ExitStatus] = call().flatMap(blk)
 
   def peek(param: CliParam): Option[param.Type] = args.get(param.param).toOption
-  
-  def preview(param: CliParam)(default: Option[param.Type] = None): Try[param.Type] =
-    args.get(param.param).toOption.orElse(default).ascribe(MissingParam(param))
+
+  def preview(param: CliParam)(default: Option[param.Type] = None): Try[param.Type] = {
+    val result = args.get(param.param)
+    if(default.isDefined) result.recover {
+      case _: exoskeleton.MissingArg => default.get
+    } else result
+  }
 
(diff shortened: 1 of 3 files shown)


src/core/cli.scala
```diff
@@ -220,9 +220,13 @@ class Cli(val stdout: java.io.PrintWriter,
...
 
   def peek(param: CliParam): Option[param.Type] = args.get(param.param).toOption
-  
-  def preview(param: CliParam)(default: Option[param.Type] = None): Try[param.Type] =
-    args.get(param.param).toOption.orElse(default).ascribe(MissingParam(param))
+
+  def preview(param: CliParam)(default: Option[param.Type] = None): Try[param.Type] = {
+    val result = args.get(param.param)
+    if(default.isDefined) result.recover {
+      case _: exoskeleton.MissingArg => default.get
```
propensive, line 227: Is the difference that we only recover `MissingArg` exceptions with the default significant? I guess they could be some sort of validation exception instead, in which case we don't want to lose that error?

odisseus: I assume that `exoskeleton.MissingArg` is thrown if and only if the expected parameter is entirely missing. This is the only case in which the default value should be used instead. All other kinds of errors, such as invalid parameter value or bad command syntax, should not be handled by this method.

odisseus: By the way, the error message for `InvalidArgValue` provided by Exoskeleton seems to mix up the name and value of the parameter:

```
exoskeleton.InvalidArgValue: the value '-m/--module' is not valid for the parameter --foo/bar
```

propensive: Can we convert it to a `BadParamValue` so that we don't have to deal with non-`FuryException` exceptions in the recovery pattern match? I would like to make `FuryException` sealed one day to ensure that we handle every sort of exception.

odisseus: That's definitely possible, but the conversion will have to be done in every method that uses Exoskeleton.

PR #1339, opened 2020-05-05
Check that the MANIFEST.MF file is the same each time

propensive:


odisseus:
When comparing the contents of the manifest file directly, it is easy to see what exactly has changed (e. g. is it just the order of dependencies or is the file totally scrambled). How is comparing MD5 hashes better?

propensive:
It makes it possible to see whether the line endings are different. I agree that plain text would be better, but invisible characters are what's causing the current problems.

readme.md
```diff
@@ -1,7 +1,5 @@
 <img src="doc/logo/render_github.png" alt="Fury">
 
-![CI](https://github.com/propensive/fury/workflows/CI/badge.svg?branch=master)
```
odisseus, line 3: Why?

PR #1346, opened 2020-05-06
Use Scala 3 0.24

propensive:
Also remove double-run of test, in case it's not necessary


propensive:
I haven't yet seen a failure when running Scala 3 just once.

PR #1348, opened 2020-05-07
Stop the Nailgun server after Fury has received the stop command

odisseus:


propensive (approved):

propensive:
This was near enough identical to the fix I just tried to push to this branch (which I didn't expect to already exist...)

PR #1353, opened 2020-05-08
Don't force exit after the Nailgun server is stopped

### LINKED ISSUE #1352: 'fury restart' stops server; but doesn't restart it
Running 'fury restart' shuts down nailgun, but does not restart it.
#1348 exits early after shutting down, ngServer never called. 

➤ ps -e | grep fury    
➤ fury start                 
➤ ps -e | grep fury                                                                                                 
  47239 pts/0    00:00:00 fury
➤ fury restart                   
(...waited a little bit)                                                                                   
➤ ps -e | grep fury

odisseus:
The [`exit`](https://www.tldp.org/LDP/abs/html/exit-status.html) command terminates the whole script, not just the function. Also, if exit code 230 means that Fury is not running, then this outcome is as good as success for the stop command. In particular, it is not unreasonable to "restart" Fury even if it wasn't running.

Fixes #1352.

### DIFF
--- a/etc/fury
+++ b/etc/fury
@@ -54,8 +54,8 @@ coursier() {
 stopFury() {
   nailgun "${FURY_MAIN}" "$$" "stop" 2> /dev/null
   case "$?" in
-    0)   nailgun ng-stop && exit 0 ;;
-    230) echo "Fury is not running." && exit 1 ;;
+    0)   nailgun ng-stop ;;
+    230) echo "Fury is not running." ;;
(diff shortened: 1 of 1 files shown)


propensive (changes requested):
.

propensive (changes requested):
I think your changes mean the `230` case is consumed, and it appears as a success. We can remove `exit 0` though.

propensive (approved):

PR #1357, opened 2020-05-08
See what happens if we use IPFS 0.5.0

### LINKED ISSUE #1283: Support for IPFS 0.6.x
The latest version of IPFS allows to use its HTTP API only with the POST method (see https://github.com/ipfs/go-ipfs/pull/7097). Attempting to use current builds of Fury with this version of the IPFS daemon results in exceptions like this one:

```
An unexpected error occurred:
  java.lang.RuntimeException: IOException contacting IPFS daemon.
  Trailer: null 405 - Method Not Allowed
  
      at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1900)
      at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1498)
      at io.ipfs.api.IPFS.get(IPFS.java:682)
      at io.ipfs.api.IPFS.retrieve(IPFS.java:675)
      at io.ipfs.api.IPFS.retrieveAndParse(IPFS.java:637)
      at io.ipfs.api.IPFS.version(IPFS.java:585)
      at io.ipfs.api.IPFS.<init>(IPFS.java:71)
      at io.ipfs.api.IPFS.<init>(IPFS.java:47)
```

This change doesn't require any changes in the Fury code, but the [java-ipfs-http-client](https://github.com/ipfs-shipyard/java-ipfs-http-client) library will have to be updated as soon as the new approach is implemented (see https://github.com/ipfs-shipyard/java-ipfs-http-client/issues/157).

propensive:


propensive:
I'm not sure this passing actually proves anything, as we've probably avoided dependencies on IPFS in our tests...

odisseus:
Fury now supports IPFS 0.6.0, so this pull request is not needed anymore.

PR #1361, opened 2020-05-09
added sorted reporter

### LINKED ISSUE #1351: produce "stable" reporter
The output from builds sometimes changes order. We should output a list of compiled modules once they all complete, sorted.

veyga:
Added a "stable" reporter, which displays compile results in sorted order according to module name. "Sorted" felt like a more appropriate name than stable. Test added to the integration suite.

resolves #1351


src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+  def report(graph: Target.Graph,
+             theme: Theme,
+             multiplexer: Multiplexer[ModuleRef, CompileEvent])(implicit log: Log)
+            : Unit = {
+    val interleaver = new Interleaver(3000L)
+    multiplexer.stream(50, Some(Tick)).foreach{logEvent}
   }
 }
+
+object SortedLinearReporter extends LinearReporter("sorted") {
```
propensive: Can we call it `summary`?

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+  import collection.mutable.{PriorityQueue => PQ}
+  type MCE = ModuleCompileEvent
```
propensive: Not to criticize this in general, but (for better or worse) we don't use any abbreviations anywhere in the code, so I'd like to keep that consistent. Can you delete these lines and write these types out in full?

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+  private val moduleOrd: Ordering[MCE] = (a: MCE, b: MCE) => (b.ref.id) compareTo (a.ref.id)
+
+  def report(graph: Target.Graph,
+             theme: Theme,
+             multiplexer: Multiplexer[ModuleRef, CompileEvent])(implicit log: Log)
+            : Unit = {
+    val (started, stopped, otherEvts) = (PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd))
+    val interleaver = new Interleaver(3000L)
```
propensive: This, I think, is an experiment which became redundant, so unless it's referred to anywhere, I think we can just get rid of this line.

test/passing/sorted-report/check
```diff
@@ -0,0 +1,31 @@
...
+Starting compilation of module test-project/module-b
+Starting compilation of module test-project/module-c
+Starting compilation of module test-project/module-d
+Starting compilation of module test-project/module-e
+Successfully compiled module test-project/app
+Successfully compiled module test-project/module-a
+Successfully compiled module test-project/module-b
+Successfully compiled module test-project/module-c
+Successfully compiled module test-project/module-d
+Successfully compiled module test-project/module-e
```
propensive: So, the summary message I would like to see would just be a single line:
```
Successfully compiled test-project/module-a, test-project/module-b, test-project/module-c, test-project/module-d, test-project.module-e
```
```

File: test/passing/sorted-report/script [hunk 5]
```diff
@@ -0,0 +1,40 @@
...
+mkdir src/module-c
+echo "object ModuleC{ val cVal = 'C' }" > src/module-c/ModuleC.scala 
+fury source add -s src/module-c
+fury module add -n module-d -c scala/compiler
+mkdir src/module-d
+echo "object ModuleD{ val dVal = 'D' }" > src/module-d/ModuleD.scala 
+fury source add -s src/module-d
+fury module add -n module-e -c scala/compiler
+mkdir src/module-e
+echo "object ModuleE{ val eVal = 'E' }" > src/module-e/ModuleE.scala
```
propensive: If it's easier (too late now, I guess...) we can just create these `.scala` files in the test directory.

src/core/reporter.scala
```diff
@@ -20,7 +20,7 @@ import fury.strings._, fury.model._, fury.utils._
 
 object Reporter {
   implicit val parser: Parser[Reporter] = unapply(_)
-  val all: List[Reporter] = List(GraphReporter, InterleavingReporter, LinearReporter, QuietReporter)
+  val all: List[Reporter] = List(GraphReporter, InterleavingReporter, UnsortedLinearReporter, SortedLinearReporter, QuietReporter)
```
propensive: Rename these (back) to `LinearReporter` and `SummaryReporter`.

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+object SortedLinearReporter extends LinearReporter("sorted") {
+  import collection.mutable.{PriorityQueue => PQ}
+  type MCE = ModuleCompileEvent
+  private val moduleOrd: Ordering[MCE] = (a: MCE, b: MCE) => (b.ref.id) compareTo (a.ref.id)
+
+  def report(graph: Target.Graph,
+             theme: Theme,
+             multiplexer: Multiplexer[ModuleRef, CompileEvent])(implicit log: Log)
+            : Unit = {
+    val (started, stopped, otherEvts) = (PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd))
```
propensive: I think all we need here is a `mutable.Set[ModuleRef]` which collects the modules when they're complete.

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+    val interleaver = new Interleaver(3000L)
+    multiplexer.stream(50, Some(Tick)).foreach {
+      case e: StartCompile => started enqueue e
+      case e: StopCompile => stopped enqueue e
+      case d: DiagnosticMsg => otherEvts enqueue d
```
propensive: So, I would a) ignore `StartCompile` messages, b) record `StopCompile` module refs in the set, and c) probably record whether any `DiagnosticMsg` events are received, and `log.note` them instead of `log.info` when they happen. This way, they will appear in the global log, but won't pollute the output if they arrive in a slightly different order sometimes.

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
+      case p: Print => otherEvts enqueue p
+      case _ => ()
+    }
+
+    def logSorted(pq: PQ[MCE]): Unit = pq.headOption match {
+      case Some(e) => logEvent(pq.dequeue); logSorted(pq)
+      case None => ()
+    }
```
propensive: I'm not sure I understand what this is doing... but I don't think we need it anyway. ;)

src/core/reporter.scala
```diff
@@ -47,22 +47,58 @@ object GraphReporter extends Reporter("graph") {
...
 }
 
-object LinearReporter extends Reporter("linear") {
-  def report(graph: Target.Graph,
-             theme: Theme,
-             multiplexer: Multiplexer[ModuleRef, CompileEvent])(implicit log: Log)
-            : Unit = {
-    val interleaver = new Interleaver(3000L)
-    multiplexer.stream(50, Some(Tick)).foreach {
+abstract class LinearReporter(name: String) extends Reporter(name){
```
propensive: I *think* (but will trust your judgement) after the changes below, that extracting out the abstract interface isn't worth it. I don't think the extra layer will give us much deduplication of code.

propensive (changes requested):
Great to see this PR in the queue, @astefanich! There are a load of comments, but don't be too discouraged by them—I'm just particularly attentive to first-time contributors, and I'm really happy to see the contribution. There should be no problem merging it with the changes. ;)

veyga:
I set up main.yml to run on my fork. It took a couple of tries for them all to pass, as something random seemed to fail every time, even without me editing anything. Is this to be expected?

propensive:
Yes, there are "random failures" more often than I'd like, but I hope that this PR will fix some of them! Did you happen to notice if the failures looked "significant"? We often see log messages appearing in a slightly different order, and sometimes the "Starting BSP..." messages appear.

Doing an exact character-for-character check on the output is one reason these tests are such high-maintenance...

src/core/reporter.scala
```diff
@@ -47,55 +47,35 @@ object GraphReporter extends Reporter("graph") {
...
             : Unit = {
-    val (started, stopped, otherEvts) = (PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd))
-    val interleaver = new Interleaver(3000L)
+    val moduleStrings = collection.mutable.TreeSet[String]()
     multiplexer.stream(50, Some(Tick)).foreach {
-      case e: StartCompile => started enqueue e
-      case e: StopCompile => stopped enqueue e
-      case d: DiagnosticMsg => otherEvts enqueue d
-      case p: Print => otherEvts enqueue p
+      case StopCompile(ref, true) => moduleStrings += ref.id
```
propensive: We would avoid using `String`s for as long as possible, here. The reason comes a bit later...

src/core/reporter.scala
```diff
@@ -47,55 +47,35 @@ object GraphReporter extends Reporter("graph") {
...
-  private val moduleOrd: Ordering[MCE] = (a: MCE, b: MCE) => (b.ref.id) compareTo (a.ref.id)
-
+object SummaryReporter extends Reporter("summary") {
   def report(graph: Target.Graph,
              theme: Theme,
              multiplexer: Multiplexer[ModuleRef, CompileEvent])(implicit log: Log)
             : Unit = {
-    val (started, stopped, otherEvts) = (PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd), PQ[MCE]()(moduleOrd))
-    val interleaver = new Interleaver(3000L)
+    val moduleStrings = collection.mutable.TreeSet[String]()
```
propensive: Keep the type as `ModuleRef` and call it `moduleRefs`.

src/core/reporter.scala
```diff
@@ -47,55 +47,35 @@ object GraphReporter extends Reporter("graph") {
...
-    def logSorted(pq: PQ[MCE]): Unit = pq.headOption match {
-      case Some(e) => logEvent(pq.dequeue); logSorted(pq)
-      case None => ()
-    }
-
-    logSorted(started)
-    logSorted(stopped)
-    logSorted(otherEvts)
+    val summary = moduleStrings.foldLeft(new StringBuilder("Succesfully compiled "))((sb,m) => sb.append(m + ", "))
+    log.info(summary.dropRight(2).toString)
```
propensive: We would never bother with anything so performance-conscious just for a log message. (FWIW, the `StringBuilder` is mutable anyway, so there's no need to fold over the reference to it—with this approach you could just use `foreach` on the `moduleStrings`.) The easy way would be to use `moduleStrings.mkString(", ")` and then you can avoid the `dropRight` on L78.

However, our `log.info` accepts `UserMsg` as input in addition to `String`. `UserMsg` provides rich colored output for terminals that support it (all of them these days). You can construct one with `msg"""Text in here with $substitutions."""`.

In this example, I can't remember if it works, but you could try substituting `moduleRefs.to[List]` in here. If it doesn't, you'll have to make them into a list. Try something like this:
```
moduleRefs.to[List].foldLeft(msg"") { case (ms, m) => msg"$ms, $m" }
```

src/core/reporter.scala
propensive: `ModuleRef`s might have an `Ordering` instance, but if they don't just use a plain `mutable.Set` and sort later.

test/passing/summary-report/check
```diff
@@ -15,17 +15,5 @@ e36c63fb9ddb0d9027f4e634bd6440182ffdf4e338677debeda2634d3a81ae5a
...
-Starting compilation of module test-project/module-d
-Starting compilation of module test-project/module-e
-Successfully compiled module test-project/app
-Successfully compiled module test-project/module-a
-Successfully compiled module test-project/module-b
-Successfully compiled module test-project/module-c
-Successfully compiled module test-project/module-d
-Successfully compiled module test-project/module-e
-A B C D E
+Successfully compiled test-project/module-a, test-project/module-b, test-project/module-c, test-project/module-d, test-project/module-e
```
propensive: The tests always run with the `nocolor` reporter, so this shouldn't have to change once you start using `UserMsg`. If you have lost the colorful output, it's probably because the integration tests didn't clean up after themselves, and you'll need to run,
```
fury config set --theme full
```

propensive (changes requested):

src/core/reporter.scala
veyga: ah, I didn't realize notice that.

veyga:
I re-ran the build file on my fork, and they passed after a few tries:
https://github.com/astefanich/fury/actions/runs/100410616

Since the tests are a little unpredictable, this is probably easier than making a new commit/PR every time I want to rerun.

propensive:
@odisseus is working on a fix for some of the test instability. I can usually check the error output before merging to check whether the test errors are real breakages or not, at least until we have more stability. One other alternative would be to run failed tests twice to see if they work the second time, but I think that just kicks the can down the road a bit...

propensive:
Ah, there's one more thing to do here... we should update most of the tests to use your sorted reporter! That would fix the one current test failure. :)

propensive:
Give me a nod when it's ready to merge, @astefanich!

veyga:
ready sir

propensive:
Very happy to have this in Fury! Thank you @astefanich, especially given all the polishing I was asking for along the way...!

PR #1362, opened 2020-05-09
guard for attempting to add duplicate repository names

odisseus:


odisseus:
Adding a test broke another test, and renaming the repository in the new test fixed the broken one. That's suspicious...

propensive (approved):
Great. I think there are a few places we need this check. I was wondering whether it was worth defining an extension method to `SortedSet` called `insert` which would guarantee uniqueness, so that every time we really want to insert, we're guaranteed an error if it fails.

Issue #1370, opened 2020-05-10
Remotes with equal relative paths are treated as if they were the same remote


odisseus:
When the remote (which used to be called a repo) is fetched, the path where it will be stored is determined by the hash of its `ref`. The ref is just a string which can be an encoded web address (e. g. `gh:propensive/fury`) or a local directory (e. g. `../exoskeleton`). In the latter case, repos defined by a relative path might map to the same directory even if that path is resolved to different absolute paths by the layouts in which they are used.

This problem is especially visible in the integration tests, because a few of them use repositories located at `../library` (relative to the base directory of each respective test). This leads to inconsistent messages "Cloning repository at...", and makes the success of seemingly independent tests depend on the order in which they are executed. This is probably also the cause of occassional corruption of cached repositories.

propensive:
That's some impressive investigative work! Would resolving to absolute paths be a fix, given that the full path may be different for different runs on GitHub Actions?

odisseus:
We will have to use the absolute path when generating the name for the cache directory. I don't think it is needed to alter the definition of the remote or the way remotes are stored in the layer.

PR #1376, opened 2020-05-11
Made the name of repo checkout directory depend on the layout base path

odisseus:
This pull request is related to the issue #1370.

### DIFF
--- a/src/core/remote.scala
+++ b/src/core/remote.scala
@@ -56,7 +56,14 @@ object Remote {
 }
 
 case class Remote(ref: String) {
-  def path(layout: Layout): Path = Installation.reposDir / ((layout.env.workDir.flatMap(d => local(Path(d))), ref).digest[Md5].encoded)
+  def path(layout: Layout): Path = {
+    val localSource = for {
+      base <- layout.env.workDir
+      foo <- local(Path(base))
+    } yield foo
+    val dirName = (localSource -> ref).digest[Md5].encoded
+    Installation.reposDir / dirName
+  }
 
   def equivalentTo(remote: Remote): Boolean = remote.simplified == simplified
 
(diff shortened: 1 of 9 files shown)


src/core/remote.scala
```diff
@@ -105,24 +104,39 @@ case class Remote(ref: String) {
...
-  def https: String = simplified match {
+  def local(baseDir: Path): Option[String] = {
+    object ExistingDirectory{
+      def unapply(p: String): Option[Path] = {
+        val absolutePath = baseDir.resolve(Path(p)).absolutePath().get
+        if(absolutePath.directory) Some(absolutePath) else None
+      }
+    }
+    ref match {
+      case r"git@.*" => None
```
propensive, line 127: I'm not sure it's guaranteed to be `git@` for all repositories.

src/core/remote.scala
```diff
@@ -105,24 +104,39 @@ case class Remote(ref: String) {
...
+      case r"https://.*" => None
+      case ExistingDirectory(dir) => Some(str"file://${dir}")
+      case _ => None
+    }
+  }
+
+  def https: Option[String] = simplified match {
```
propensive: I think the `only` extension method could replace `match` and the `wildcard`. There are lots of examples of its usage in `ids.scala`.

propensive (commented):
This can be merged when you think it's ready, and the tests are passing. Looks like there's only one simple line missing from one of the test checks.

src/core/remote.scala
odisseus: Refactored

propensive (approved):

src/core/remote.scala
odisseus: You are right, Git [supports](https://www.git-scm.com/docs/git-clone#_git_urls) many different address formats and transport protocols. But the assumption that a "git@" prefix means SSH and vice versa already existed in the code before this pull request.

PR #1379, opened 2020-05-11
Simplify hello-world

propensive:


odisseus:
This test used to be an example of setting up a Fury project from scratch. If it is replaced with a ready-made layer, the test will lose its educational value.

propensive:
Yes, maybe it wasn't the best choice to give this "simplification" treatment to, given that I want most of the integration tests to be written this way and could have chosen any... But I'd like to have more integration tests which do some of the basics, and potentially fail at those.

Issue #1350, opened 2020-05-07
allow dependencies on local JARs


propensive:
We should be able to point to a file, rather than Maven coordinates to specify a binary. It shouldn't be possible to publish a layer containing a local JAR, unless the JAR is inside a repo.

veyga:
should a local jar need only be resolved at compile time?
One can add remote binaries, even if they don't resolve on maven central. They show up on 'fury binary list' but of course don't compile/resolve. Is this the desired behavior, or should jars be validated on add?

Is the git constraint simply that it is within any git repository? Or is there some intersection with fury.

Issue #544, opened 2019-09-26
adding binaries should be synchronous


propensive:
When adding a binary, it's not resolved until attempting the build. This should be done when the binary is added, and any error messages (such as "binary not found") should be displayed at the point of adding. If it's not found, it shouldn't be added.

Currently, if a binary is not found then the error is not handled, and a stack trace is generated.

This shouldn't happen if the user is offline, though.

odisseus:
What if the user attempts to add a binary that is temporarily unavailable – e. g. during a network outage?

propensive:
Require them to add `--force` maybe?

propensive:
I think it might depend on how easily we can disambiguate between a network failure and a certain message that the binary is not found.

odisseus:
The best option is probably an interactive prompt asking the user what to do, and maybe a force flag to add the reference no matter what.

propensive:
Though we don't do anything interactively yet, and this doesn't seem to be a good enough reason to start.

PR #1382, opened 2020-05-12
add locals jars/ validate binaries on import

veyga:
Added an option on 'fury binary add' to allow both remote and local jars. Both kinds are validated on initial import; local jars according to their absolute path.
Added a path attribute which prints in table, to denote where imported jars are save on host filesystem. 
Creating a new type to distinguish between remote and local jars ripped through client code and the type checker went AWOL, so I kept a single 'Binary' type.
#1350 
#544


propensive:
This sounds awesome, and I'm impressed you've managed to find your way around! I think @szymonrodziewicz was asking for this recently...

I haven't looked in detail yet, but can we have the JARs referred to using the same command syntax that we use for resources? There's a problem with absolute paths because it makes the builds non-portable, and there's a potential security issue with someone publishing a build with a reference to `/home/enemy/secrets.jar` and then doing something evil with it in an application module. Something similar to the way we specify resources would allow us to choose JAR files stored inside Git repos or inside the working directory.

I'll try to take a closer look tonight.

src/core/args.scala
```diff
@@ -29,14 +29,15 @@ object Args {
...
   val AliasArg = CliParam[AliasCmd]('a', 'alias, "specify a command alias")
   val ActionArg = CliParam[String]('A', 'action, "specify a permission action")
   val BinaryArg = CliParam[BinaryId]('b', 'binary, "specify a binary")
   val ColumnArg = CliParam[String]('c', 'column, "specify the column")
   val PartialBinSpecArg = CliParam[PartialBinSpec]('b', 'binary, "specify a binary")
-  val BinaryNameArg = CliParam[BinaryId]('n', 'name, "specify the name for the binary")
-  val BinSpecArg = CliParam[BinSpec]('b', 'binary, "specify a binary dependency")
+  val BinaryNameArg = CliParam[BinaryId]('n', 'name, "specify an alias for the binary")
+  val BinSpecArg = CliParam[BinSpec]('b', 'binary, "specify a remote binary dependency")
+  val LocalJarFileArg = CliParam[JarFile]('l', 'local, "specify a local binary dependency")
```
propensive: Let's use `-f` and `--file` for this parameter, which is what we have done elsewhere.

src/core/binary.scala
```diff
@@ -28,19 +28,48 @@ object Binary {
   implicit val msgShow: MsgShow[Binary] = v => UserMsg(_.binary(v.spec))
   implicit val stringShow: StringShow[Binary] = b => b.id.key
   implicit def diff: Diff[Binary] = Diff.gen[Binary]
+  implicit val pure = (s: String) => if(s.isEmpty) None else Some(s)
```
propensive: This gives up too much typesafety, unfortunately, and is generally not recommended. Can you delete this line and just wrap everything in `Some` where it's needed?

propensive: Oh, also, in general implicits should *always* have an explicit return type.

veyga:
What if I simply copied the jar into something like .fury/.lib in the working directory, or perhaps a directory in ~/.cache/fury/bins. That way the absolute path is only used for initial validating/locating the file. Then on 'fury binary remove' it can be evicted. 

I haven't accounted for the git constraint yet. Is the validation simply checking that the jar is contained with any .git directory/subdirectory when its initially added?

src/core/binary.scala
```diff
@@ -28,19 +28,48 @@ object Binary {
...
-case class Binary(id: BinaryId, binRepo: BinRepoId, group: String, artifact: String, version: String) {
+case class Binary(
+  id: BinaryId, 
+  binRepo: BinRepoId, 
+  group: String = Binary.NA, //TODO add option wrappers? it ripples through client code :(
+  artifact: String = Binary.NA,
+  version: String = Binary.NA,
+  path: Option[Path] = None
+) {
   def spec = str"$group:$artifact:$version"
```
propensive: I think this will need to be cleaned up. The `NA` value isn't typesafe enough. I think there's a real risk of methods assuming that it's a reasonable value. I think that a more drastic redesign of the `Binary` type is necessary. In general, trying to maintain a not-quite-perfect representation of the model while the rest of the code is changing causes proliferation of bugs. So in general, a priority is put on making the model definitions an exact representation of only the states we want to represent, as it saves problems down the line, at the expense of a greater up-front cost.

src/module/module.scala
```diff
@@ -323,15 +324,19 @@ case class BinaryCli(cli: Cli)(implicit log: Log) {
     call         <- cli.call()
     project      <- optProject.asTry
     module       <- project.modules.findBy(moduleId)
-    binSpecArg   <- call(BinSpecArg)
+    targetBin    <- call.exactlyOne(BinSpecArg, LocalJarFileArg)
```
propensive: Glad you found this method!

src/model/ids.scala
```diff
@@ -677,6 +678,19 @@ object BinSpec {
 
 case class BinSpec(string: String)
 
+object JarFile {
+  implicit val msgShow: MsgShow[JarFile] = b => UserMsg(_.binary(b.path))
+  implicit val stringShow: StringShow[JarFile] = _.path
+  implicit val diff: Diff[JarFile] = (l, r) => Diff.stringDiff.diff(l.path, r.path)
+  implicit val parser: Parser[JarFile] = unapply(_)
+  implicit val keyName: KeyName[JarFile] = () => msg"local jar file"
+  implicit val jarPath = (bin: JarFile) => Path(bin.path)
```
propensive: This implicit should be removed for the same reasons.

test/passing/binary-validation/check
```diff
@@ -0,0 +1,22 @@
...
+0
+Either the --binary (-b) or --local (-l) parameter must be specified.
+Binary file could not be found: /etc/dummy.jar
+0
+The binary commons-io already exists.
+commons-io
+commons-math3
+Starting the BSP launcher...
+Opening a BSP server connection
+BSP server is listening for incoming connections
```
propensive, line 20: I think these three lines only appear if the test is run before Fury has started a BSP connection. They probably won't be shown if the test is run in CI.

test/passing/binary-validation/description
```diff
@@ -0,0 +1,3 @@
+Local jars should be includable in a module.
+Local and remote binaries should be validated for syntax 
+and availability when they are added to a module.
```
propensive, line 3: This file was meant to be a one-liner, though I guess this just prints out three lines instead of one in the output... ;)

test/passing/binary-validation/script
```diff
@@ -0,0 +1,22 @@
...
+echo $?
+fury binary add -l /etc/dummy.asdf
+fury binary add -l /etc/dummy.jar
+fury binary add -l ./bins/commons-io.jar
+echo $?
+fury binary add -l ./bins/dummy.jar -n commons-io
+fury binary list --raw
+fury source add -s src
+fury build run --output summary
+echo $?
```
propensive: This all looks pretty solid. It's not too important, but I'm hoping to make these test scripts shorter (and faster) by running the first few commands (let's say lines 1-7 in this file) in the test directory, deleting them from the script file, and then running a `fury layer commit` to add it to the Git repository. PR #1379 is a WIP example of this change.

propensive (changes requested):
So, firstly, well done for managing to navigate around such a variety of different parts of the code!

The most important change needed will be find a precise representation of the binary. One possibility would be the following:
- restore `Binary` to its previous form
- add a `jarFiles` `SortedSet[JarFile]` to `Module` which accurately represents a JAR file which may be stored in a repository, or as a local file.
- introduce a common supertrait to `Binary` and `JarFile` so that in the places we need a common representation (e.g. the tabular list) we can just give it `binaries ++ jarFiles` and it will display the appropriate information for each.

I would like to reuse the representation of `Resource` if possible, but it might be easier to create new types. One possibility to explore would be to literally use `Resource`s in their current form. They exist only to included extra files in produced JAR files, but it would be a matter of a boolean flag to determine if they (as JAR files) should also be included on the compile classpath. I'm not sure this would be our ultimate solution, but I think it would get us closer.

Modifying the data that's stored in the layer files is tricky, by the way: adding new fields is generally safe, but removing or renaming any requires a new version number, and some short migration code to be written to ensure that layer files remain forwards-compatible.

Let me know if you need any more pointers, and thanks for persevering!

propensive:
> What if I simply copied the jar into something like .fury/.lib in the working directory

I think this doesn't work in general. The build that gets run should always reflect the current state, so if that JAR file changes (especially if it's changed by Git), there should be no opportunity for it to accidentally keep using the old cached version.

veyga:
What about hashing jar/files external to the layer to ipfs?

Most commands will display an absolute path if you give it one. Pretty much everything operates on a path relative to the layer. But if you reference a directory or a file that's outside the layer, the full path gets resolved and displayed in a 'list' command as well as on the build file in ipfs.

Its probably not an issue if you're referencing a directory, since it will look like this:
tmp/someSudDir//**

But if you add a file, namely a resource file or a jar, it gets displayed in full, with glob suffixed to the file.
tmp/someSudDir//sometext.txt//**

Since both resources and binaries paths need to concealed somehow,
maybe encrypt the path, create a separate ipfs hash for each resource external to the layer, or not displaying it at all?
 I think packaging a fat jar would remove the need to resolve paths. And unless its the importer of a build file is on a share filesystem, they could probably only resolve the resource with something like ipfs.

src/core/binary.scala
```diff
@@ -24,23 +24,40 @@ import kaleidoscope._
 import scala.collection.immutable._
 import scala.util._
 
+//TODO rename me
+trait UniqueBinary {
```
veyga, line 28: ideas on better name, if we can't change the name of the other Binary

src/module/module.scala
```diff
@@ -313,25 +316,35 @@ case class BinaryCli(cli: Cli)(implicit log: Log) {
...
     cli          <- cli.hint(BinSpecArg, binSpecs)
+    cli          <- cli.hint(JarPathArg)
     cli          <- cli.hint(ProjectArg, layer.projects)
     optProjectId <- ~cli.peek(ProjectArg).orElse(layer.main)
     optProject   <- ~optProjectId.flatMap(layer.projects.findBy(_).toOption)
     cli          <- cli.hint(ModuleArg, optProject.to[List].flatMap(_.modules))
     moduleId     <- cli.preview(ModuleArg)(optProject.flatMap(_.main))
     cli          <- cli.hint(BinaryNameArg)
-    cli          <- cli.hint(BinaryRepoArg, List(RepoId("central")))
+    cli          <- cli.hint(BinaryRepoArg, List(RepoId("mvn_central")))
```
veyga: leave as central?

src/core/source.scala
```diff
@@ -129,3 +131,24 @@ case class LocalSource(dir: Path, glob: Glob) extends Source {
   def base(checkouts: Checkouts, layout: Layout): Try[Path] = Success(layout.baseDir)
 }
 
+case class JarResource(id: BinaryId, repoId: RepoId, dir: Path, glob: Glob) extends Source with UniqueBinary {
+  //TODO define appropriate key for jar and other resources
```
veyga, line 135: this key hides the absolute path. I think this plus the other resources should implement some sort of concealment, since if you give it an absolute path, it will display it. However this would affect exists builds and the parsing of keys.

propensive:
@odisseus Do you think you would be able to give @astefanich any feedback on this? I'm tied up for the next day or so... (@astefanich sorry!)

odisseus:
Okay, I'll have a look.

veyga:
@propensive all good; no rush

src/core/binary.scala
```diff
@@ -24,23 +24,40 @@ import kaleidoscope._
...
+
 object Binary {
   implicit val msgShow: MsgShow[Binary] = v => UserMsg(_.binary(v.spec))
   implicit val stringShow: StringShow[Binary] = b => b.id.key
   implicit def diff: Diff[Binary] = Diff.gen[Binary]
 
-  def apply(id: Option[BinaryId], service: BinRepoId, binSpec: BinSpec): Try[Binary] =
-    binSpec.string.only {
-      case r"$group@([\.\-_a-zA-Z0-9]*)\:$artifact@([\.\-_a-zA-Z0-9]*)\:$version@([\.\-\+_a-zA-Z0-9]*)" =>
+  def apply(id: Option[BinaryId], service: BinRepoId, binSpec: BinSpec)(implicit log: Log): Try[Binary] ={
```
odisseus, line 42: Is the log actually used in this method?

src/core/binary.scala
odisseus: What is the purpose of this trait?

src/module/module.scala
odisseus: The value of this argument must be either an URL of the repository (such as `https://repo1.maven.org/maven2`) or one of the standard repository names [recognized](https://get-coursier.io/docs/other-repositories) by Coursier. The name "central" corresponds to Maven Central Repository, which is a sensible default value.

src/core/binary.scala
veyga: its passing implicitly in the call to paths()

src/core/source.scala
odisseus: I don't understand why the absolute path to a binary should be hidden. The user has a choice whether to specify the path to the binary as an absolute or a relative one. Each option has its benefits and drawbacks, and both should be supported by Fury.

src/core/binary.scala
```diff
@@ -24,23 +24,40 @@ import kaleidoscope._
...
-case class Binary(id: BinaryId, binRepo: BinRepoId, group: String, artifact: String, version: String) {
-  def spec = str"$group:$artifact:$version"
-  def paths(implicit log: Log): Try[List[Path]] = Coursier.fetch(this)
+case class Binary(id: BinaryId, binRepo: BinRepoId, group: String, artifact: String, version: String) extends UniqueBinary {
+  val spec = str"$group:$artifact:$version"
+  def paths(implicit log: Log): Try[List[Path]] = binRepo match {
+    case BinRepoId.Central => Coursier.fetch(this).recoverWith {
+        case OfflineException() => ~List[Path]()
+    }.orElse(Failure(DownloadFailure(spec)))
+    case _ => Failure(UnknownBinaryRepository(binRepo))
```
odisseus: Does this mean that an attempt to use a repository other than Maven Central will fail? If so, Fury will be unable to build itself.

src/core/binary.scala
```diff
@@ -24,23 +24,40 @@ import kaleidoscope._
...
   val Jmh = Binary(BinaryId("jmh-core"), BinRepoId.Central, "org.openjdk.jmh", "jmh-core", "1.21")
 }
 
-case class Binary(id: BinaryId, binRepo: BinRepoId, group: String, artifact: String, version: String) {
-  def spec = str"$group:$artifact:$version"
-  def paths(implicit log: Log): Try[List[Path]] = Coursier.fetch(this)
+case class Binary(id: BinaryId, binRepo: BinRepoId, group: String, artifact: String, version: String) extends UniqueBinary {
+  val spec = str"$group:$artifact:$version"
+  def paths(implicit log: Log): Try[List[Path]] = binRepo match {
+    case BinRepoId.Central => Coursier.fetch(this).recoverWith {
```
odisseus: Use [`Try.recover`](https://www.scala-lang.org/api/2.12.10/scala/util/Try.html#recover[U%3E:T](pf:PartialFunction[Throwable,U]):scala.util.Try[U]) to avoid the need to wrap the fallback value into a Try.

src/core/module.scala
```diff
@@ -50,7 +51,12 @@ case class Module(id: ModuleId,
                   optDefs: SortedSet[OptDef] = TreeSet(),
                   deterministic: Boolean = false) {
 
-  def allBinaries: SortedSet[Binary] = if(kind == Benchmarks) binaries + Binary.Jmh else binaries
+  private def remoteBinaries: SortedSet[Binary] = if (kind == Benchmarks) binaries + Binary.Jmh else binaries
+  private def localJars: SortedSet[JarResource] = jars.collect { case src: JarResource => src }
+  def allBinaries: SortedSet[UniqueBinary] = {
+      implicit val binOrd: Ordering[UniqueBinary] = (a,b) => (a.id.key) compareTo (b.id.key)
```
odisseus: I think [`Ordering.by`](https://www.scala-lang.org/api/2.12.10/scala/math/Ordering$.html) can be used here.

src/core/binary.scala
odisseus: I think this trait should be named `Binary`, and the old `Binary` should be renamed to something like `ManagedBinary`. The renaming should be done in a separate pull request though, because this one is quite large already. @propensive, what do you think about this?

src/core/source.scala
```diff
@@ -129,3 +131,24 @@ case class LocalSource(dir: Path, glob: Glob) extends Source {
   def base(checkouts: Checkouts, layout: Layout): Try[Path] = Success(layout.baseDir)
 }
 
+case class JarResource(id: BinaryId, repoId: RepoId, dir: Path, glob: Glob) extends Source with UniqueBinary {
```
odisseus, line 134: Why does this class extend `Source`? The abstract class `Source` was meant to represent a directory (either local or fetched from a repository) with files that can be selected by a glob. How does a binary JAR file match this concept?

odisseus (changes requested):

src/core/source.scala
veyga: From Jon's previous post
"There's a problem with absolute paths because it makes the builds non-portable, and there's a potential security issue with someone publishing a build with a reference to /home/enemy/secrets.jar and then doing something evil with it in an application module. Something similar to the way we specify resources would allow us to choose JAR files stored inside Git repos or inside the working directory."

One can specify a binary as an absolute path or a relative one, as the full path will get resolved when one is operating inside the layer. Since its only a question of how to display the path, I was simply under the impression we shouldn't be displaying/referencing absolute paths.

src/core/binary.scala
veyga: ah. will change

veyga:
I tried to heed Jon's advice with regards to the naming and design of the binaries (not changing the existing binary, and trying to apply the resource idea to jars). 
It was mentioned in the initial issue that jars should not be able to be published unless they are contained inside a repo. This constraint has not been implemented yet, but if there is to be some intersection between jar files and repos in the future, then extending source would save some work later on, as a jar would be intersection type between the existing binary and source models.

src/core/source.scala
odisseus: This problem didn't occur to me. Unfortunately, forbidding references to absolute paths does not solve it at all: a malicious publisher could create a reference to `../../../../../../../../../../home/enemy/secrets.jar`, which in nearly all cases would be equivalent to an absolute path.

In fact, I just realized that Fury allows you to specify any directory, including `/root/.ssh`, as a resource directory, and make its contents accessible to an application module. Fortunately, an application module cannot use the network to leak these contents unless it is granted the permission to do so.

veyga: Agreed, my previous post mentions something like this. I don't believe its about forbidding references to absolute paths, but rather simply changing how they are displayed. The "key" value is used for JarResource effectively hides this. Even if type in an absolute path, it will only every display it with the ending file name, and conceal directories.

fury binary add -f ../../../../home/enemy/secrets.jar  
fury binary list
Id | BinRepo | Group | Artifact | Version | Path
--- | ----------- | -------- | ----------- | ----------| -------
secrets | local | - | - | - | local#file://**/secrets.jar

propensive:
@astefanich, I will try to take a good look at this later, and see if I can merge it. I'm probably about to merge a PR which will create some conflicts, but don't worry about them; I will fix them when I merge this.

src/core/source.scala
propensive: I think that the `Binary` type may need to be a coproduct, a bit like `Source` is, with different variants, probably something like: `RepoBinary` (for binaries stored in a linked `repo`), `LocalBinary` for local files, and `BinaryCoordinates` for references to remote Maven-style repositories. As @astefanich mentions below, I suggested trying to leave the existing `Binary` unchanged, and utilizing `Resource`s, which closely resemble `Source`s. That may still be the way to go, but I can't be sure without exploring it further. I'm not sure whether a) one approach is clearly easier than the other, b) one approach is clearly _better_ than the other, or c) whether those answers are the same.

I think my mistake was in marking this as *good first issue*. I think it's a good first issue only if you want to gain exposure to large parts of Fury's core while potentially making several wrong turns along the way... (Does that apply to you, @astefanich?) ;)

veyga:
@propensive this was good exposure! There's probably a bit more clean up to do before merging this honestly. The "resource" entity doesn't exist, but it represented by Source, so I utilized that. I think if there's a desire to add more types like LocalBinary, RepoBinary, etc. then it would be wiser to revert back to the distinction between binaries and sources. Take a look at the output for binaries as well (regarding the jars "key" and how they are displayed together in the tabulation).

propensive:
Ok, sure. I'll keep trying to give feedback where I can, though we might need to make this part of a bigger set of changes if we refactor `Source`, `Resource` and `Binary`.

src/core/source.scala
odisseus: Perhaps I just don't understand which problem you're trying to solve. 

If you just want to hide the absolute path to a JAR when it is displayed in a table (e. g. to make the output shorter), that's just a cosmetic decision which can be easily changed later. I'm not particularly fond of the way the path looks in your example, because it looks as if it contained a glob but is not (or is it?) interpreted as a glob. But the output format can be improved later, so I'm okay with it.

If you attempt to hide the "unsafe" path to make it more "safe" while keeping the original "unsafe" one in the layer, this achieves absolutely nothing in terms of security. The "unsafe" path will be still in the layer definition, available for potential abuse by the layer creator.

If you want to rewrite the "unsafe" path into a more "safe" form and store that in the layer instead of the original path, it still doesn't achieve anything in terms of security, because relative paths  are just as unsafe as absolute ones. Moreover, it might confuse the user if the rewritten path resolves in an unexpected way under different conditions (such as another base directory or another computer).

veyga: What would be a good suggestion for concealing the path? Perhaps hashing it?
If concealment is not needed then it's not needed, but my understanding is that it is.

There just doesn't seem to a clear idea how binaries should be represented in text as well as in the model (coproduct vs "resource", remote vs local vs repo, association with git). Having them as a sort of sealed trait makes the most sense to me, but the intersection between raw jar files and those contained within a repo make things awkward/redundant.

odisseus: I believe the path should be stored in the layer as it was entered by the layer creator, without concealing anything. As for the way paths are displayed, it doesn't matter too much at the moment; you might as well keep the current way of shortening them.

src/core/source.scala
odisseus: Here's what I think about the binaries of different provenance and their relationship with local and repo paths.

There are just two kinds of binary. One is the "managed binary" which is downloaded from a binary repository with the help of Coursier and is identified by its "binary coordinates", i. e. its group, artifact ID, and version. The other one is the "supplied binary" which must be present in the environment when Fury runs the build; it is identified by the path at which it can be found. Support for the latter kind is the goal of #1350.

Now as for the paths that identify the "supplied binaries", these are also divided into two kinds. The first one is the plain old filesystem path; both absolute and relative paths fall into this category. The second kind is the "repo path", such as `library:src/core` ([example](https://github.com/propensive/fury/blob/master/test/passing/platform/script#L22)). In a way, these paths are both "relative" (they are mapped to different filesystem paths depending on the system environment) and "absolute" (they are resolved to the same filesystem path regardless of the command's working directory). 

Both kinds of paths are already implemented for _sources_: `LocalSource` corresponds to the filesystem path, and `ExternalSource` to the "repo path". However, these classes represent not just paths, but _source directories_, hence they also contain data unrelated to paths (namely globs).

Here are some consequences of these differences:
* If a layer refers only to "managed binaries", its build artifacts can be published to binary repositories like Maven Central or Artifactory.
* If a layer also refers to "supplied binaries" that are stored in repos, it can be published as a Fury layer. Further restrictions may apply: e. g. it probably shouldn't be published if the repo's "remote" address is just a local path.
* If a layer contains references to "supplied binaries" at filesystem path, it should not be published at all, because those paths would be unresolvable on any other filesystem. It can still be shared though.

What @propensive called a "RepoBinary" in his previous comment is basically a "supplied binary" with a "repo path", and his "LocalBinary" is my "supplied binary" with a filesystem path.

PR #1383, opened 2020-05-13
Cleaning up representation of `Kind` in `Module`s

### LINKED ISSUE #104: module type should be serialized and modeled differently
The `Kind` type should be represented as a sealed trait, which includes the additional parameter specific to each case.

propensive:


propensive:
Kinds are now represented using a proper coproduct type in the model, and the layer files. Migration code has been added to make all this work. One consequence is that it's not possible to change the type of a module now without specifying the necessary parameters to that new type.

src/build/build.scala
```diff
@@ -364,8 +364,8 @@ case class BuildCli(cli: Cli)(implicit log: Log) {
     project      <- tryProject
     module       <- tryModule
     compilation  <- Compilation.syncCompilation(layer, module.ref(project), layout, false)
-    _            <- if(module.kind == Application) Success(()) else Failure(InvalidKind(Application))
-    main         <- module.main.ascribe(UnspecifiedMain(module.id))
+    _            <- if(module.kind.is[App]) Success(()) else Failure(InvalidKind(App))
```
odisseus: `module.kind.as[App].ascribe(InvalidKind(App))`

src/model/kinds.scala
```diff
@@ -0,0 +1,90 @@
...
+    case Lib()            => msg"lib"
+    case App(main)        => msg"app${':'}$main"
+    case Plugin(id, main) => msg"plugin${':'}$main"
+    case Compiler(spec)   => msg"compiler${':'}$spec"
+    case Bench(main)      => msg"bench${':'}$main"
+  }
+
+  implicit def stringShow: StringShow[Kind] = msgShow.show(_).string(Theme.NoColor)
+
+  def execute(kind: Kind): Boolean = kind match {
```
odisseus: I think this is a bad choice of name for a method that doesn't execute anything and doesn't even have any side effects.

src/model/kinds.scala
```diff
@@ -0,0 +1,90 @@
...
+    case App(_) | Bench(_) => true
+    case _                 => false
+  }
+}
+sealed trait Kind {
+  def as[T: ClassTag]: Option[T] = this.only { case t: T => t }
+  def is[T: ClassTag]: Boolean = this match { case t: T => true case _ => false }
```
odisseus, line 47: `this.as[T].isDefined`

odisseus:
Generally these changes look good, especially since the integration tests are passing. However, I don't like the fact that it's impossible to have underspecified or overspecified modules.

Sure, the "main class" entry isn't particularly useful for a "lib" module, but a particular module is not always going to keep the same kind. Consider a particularly long running test module:
```
fury module add -n long-running-tests -c scala/compiler -t app -M TestsSuite
```
Now suppose I'm refactoring the code of the tests in the watch mode. I don't want to wait till all the tests finish every time I make a change — I just want to check that it still compiles:
```
fury module add -n long-running-tests -t lib
```
At this point, the information about the main class is lost. When I want to enable test execution back again, I have to provide the class name again (which can be something like `org.scalatest.tools.Runner` for Scalatest). There isn't even an easy way to check what the class name was before the change (except maybe layer diff).

src/model/kinds.scala
propensive: I'll leave this as it is to avoid an allocation.

propensive:
That's a good point. There is also an unresolved issue that failing to specify the `main` parameter at the same time as the `app` type causes silent failure. That's a blocker for merging this for now.

The use case you describe is inventive, but I would have thought there would be equally inventive ways (or dumber ways) of getting the same behavior. Commenting out the test, for example, or changing a dependency. Maybe in the future we would add timeouts to running application modules.

odisseus:
Changing the module type to avoid running the tests isn't a hypothetical scenario I invented on the spot. I have done that more than once to skip execution of Fury's test modules (back when those modules still existed). The only other way to compile executable code without executing it requires moving all the code into a separate library module, and keeping the application module empty. 

Introducing timeouts for module execution will not solve this problem, because the delay is not the only reason why the user might want to skip execution: for example, the module might allocate expensive resources (such as database connections) or perform irreversible actions (e. g. sending signals over the network).

propensive:
I didn't mean that it was an unlikely scenario, just that we should find a better way of giving the user the ability to make temporary changes like this.

propensive:
Attention to @odisseus and @astefanich! This PR will upgrade the layer file format to a version which is not backwards compatible. If you start using a version of Fury which includes this PR, and write changes to a layer, those layers will not be readable by older versions of Fury, however, any old layers will remain forwards-compatible with all foreseeable future versions of Fury.

tl;dr If you use this version of Fury, there's no going back.

PR #1384, opened 2020-05-14
include printlns in summary report

veyga:
Set up to print sequentially after the compilation results line


odisseus:
You can [create empty commits](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---allow-empty) in order to trigger commit actions on a remote server.

Issue #1377, opened 2020-05-11
Issues with bloop cache and Metals


szymon-rd:
While working with stack a consisting of Fury, Metals, bloop and VSCode I would encounter issues that mainly manifested as:
 - Lack of or outdated code completion suggestions - edit: Sometimes restart of metals server would help. It occurred after working on a project for some time, and most commonly it couldn't provide information about libraries used in the project. It was the most common state - not having information about the libraries, but being able to see the project context.
 - Being unable to build a new version of code base - each build would result in compilation of the old, cached one edit: This building issue occurred pretty rarely - about once in 10 builds I think, and then persisted until I restarted fury and deleted all the building-related dirs (.fury, .bloop, .metals). I will add this to the problem description. But important part I think is that deleting .bloop file didn't help. At this point I often tried restarting fury and deleting .metals and .fury file and I think that was the operation that made it work.

I'm using MacBook Pro with macOS 10.15.4 (19E287), accessing Fury from zsh.

propensive:
Can you try something to diagnose this? I wonder if the caching issue might be a different fixed issue.

Are you able to make a deliberate breaking change to a file, attempt to recompile it, and appear to see successful compilation, reliably and repeatably?

Can you find the JSON config file in the .bloop directory for the module you're compiling, and look at the "sources" field. Is it pointing to the correct sources, or to a file in Fury's managed space?

szymon-rd:
> Is it pointing to the correct sources, or to a file in Fury's managed space?

It is pointing to correct sources in my workspace.

This building issue occurred pretty rarely - about once in 10 builds I think, and then persisted until I restarted fury and deleted all the building-related dirs (.fury, .bloop, .metals). I will add this to the problem description. But important part I think is that deleting .bloop file didn't help. At this point I often tried restarting fury and deleting .metals and .fury file and I think that was the operation that made it work. 

The code suggestion issue I was never able to fix reliably. Sometimes restart of metals server would help. It occurred after working on a project for some time, and most commonly it couldn't provide information about libraries used in the project. It was the most common state - not having information about the libraries, but being able to see the project context.

szymon-rd:
I think I reproduced the building cache issue. I had a module named webapp with classes structured within directory src/acrimony/. Then, I removed the source set from the module: 
`fury source remove -d sz:src/acrimony`
Where sz is my repository. Then I created a new folder _a/workaround_ and added this as a source. It resulted in updating the entry in bloop:
[...]

PR #1388, opened 2020-05-17
evict old class files for correct jar artifact

veyga:
Old class files are being cached which is leading to incorrect jar artifacts (see my lengthy explanation regarding #1377 ). I think simply deleting a module's classes directory when sources are removed is the easiest fix. Since we only have source paths and bloop outputs to target directories, without a constraint between source paths and full qualified class names (which is as much a scala thing as it is fury), it seems virtually impossible to find and 'target delete' old class files. 
The staging directory where jars are created never gets deleted, so any class ever included in the manifest, will forever be included in a jar. Since this directory is not addressed by any of the existing clean mechanisms, simply deleting it after every jar creation does the trick.

### DIFF
--- a/src/core/policy.scala
+++ b/src/core/policy.scala
@@ -43,7 +43,12 @@ object Policy {
   }
   
   def standardPrivileges: List[Privilege] =
-    List(Privilege(GlobalScope, Permission(ClassRef("java.util.PropertyPermission"), "scala.*", Some("read"))))
+    List(
+      Privilege(GlobalScope, Permission(ClassRef("java.util.PropertyPermission"), "scala.*", Some("read"))),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.RuntimePermission"), "accessClassInPackage.sun.misc", None)),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.RuntimePermission"), "accessDeclaredMembers", None)),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.reflect.ReflectPermission"), "suppressAccessChecks", None)),
+    )
 }
 
 case class Policy(version: Int, policy: SortedSet[Privilege] = TreeSet()) {
--- a/test/passing/class-eviction/script
+++ b/test/passing/class-eviction/script
@@ -1,0 +1,24 @@
+fury layer init
+fury project add -n scala
+fury module add -n compiler -t compiler
+fury module update -C scala-lang.org:scala-compiler:2.12.8
+fury binary add -b org.scala-lang:scala-compiler:2.12.8
+fury project add -n cache-test
+fury module add -n my-module -c scala/compiler
+fury permission require -C java.util.PropertyPermission -T test.property -A read
+fury permission require -C java.lang.RuntimePermission -T getenv.TEST1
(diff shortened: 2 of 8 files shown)


test/passing/class-eviction/script
```diff
@@ -0,0 +1,24 @@
...
+fury permission require -C java.util.PropertyPermission -T test.property -A read
+fury permission require -C java.lang.RuntimePermission -T getenv.TEST1
+fury permission require -C java.io.FilePermission -T '.content' -A write
+mkdir -p src/libA
+fury source add -s src/libA
+echo 'package libA; object LibA { val aVal = "A" }' > src/libA/liba.scala
+mkdir libB
+fury source add -s libB
+echo 'object LibB { val bVal = "B" }' >> libB/libb.scala
+fury build run -F -f out --output quiet | awk '{print $1}'
```
odisseus: Don't discard the output unless it contains variable parts, such as absolute paths or commit hashes. Even then, you should try to discard as little as possible, ideally just the variable part. Changes in the incidental output (e. g. size of the created JAR file) sometimes help spot unexpected problems.

test/passing/class-eviction/script
```diff
@@ -0,0 +1,24 @@
...
+jar -tf out/cache-test-my-module.jar | grep LibA
+jar -tf out/cache-test-my-module.jar | grep LibB
+fury source remove -s libB
+fury build run -F -f out --output quiet | awk '{print $1}'
+jar -tf out/cache-test-my-module.jar | grep LibA
```
odisseus: If you remove the `-F` flag from the build commands, you won't have to filter out the standard library classes from the archive listing, and the grep pipe will become unnecessary.

odisseus (changes requested):
Good job, but the integration test can be improved a little bit.

src/core/policy.scala
```diff
@@ -43,7 +43,12 @@ object Policy {
   }
   
   def standardPrivileges: List[Privilege] =
-    List(Privilege(GlobalScope, Permission(ClassRef("java.util.PropertyPermission"), "scala.*", Some("read"))))
+    List(
+      Privilege(GlobalScope, Permission(ClassRef("java.util.PropertyPermission"), "scala.*", Some("read"))),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.RuntimePermission"), "accessClassInPackage.sun.misc", None)),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.RuntimePermission"), "accessDeclaredMembers", None)),
+      Privilege(GlobalScope, Permission(ClassRef("java.lang.reflect.ReflectPermission"), "suppressAccessChecks", None)),
```
odisseus, line 50: Why?

test/passing/scala3/script
```diff
@@ -8,9 +8,6 @@ fury project add -n hello-world
 fury module add -n app -c scala3/compiler -t application -M HelloWorld
 fury source add -s src
 mkdir -p src
-fury permission require -C java.lang.RuntimePermission -T accessClassInPackage.sun.misc
```
odisseus, line 11: Is this related to the issue with stale source and class files?

veyga:
closes #1377

PR #1392, opened 2020-05-18
Default permissions for running Scala 3

### LINKED ISSUE #1390: grant permissions to run Scala 3 "Hello World" by default
"Hello World" compiled for Scala 3 requires some additional security permissions which it doesn't have by default.

propensive:


odisseus:
Why couldn't these permissions be added in the test script? Are they general enough and harmless enough to grant them to every application module?

PR #1397, opened 2020-05-19
Use latest Magnolia, and ReadOnlyCaseClass where possible

### LINKED ISSUE #1385: upgrade to latest Magnolia
We should switch to using 0.16.0+ to take advantage of the faster derivations for read-only types.

propensive:
In my tests it looks like this cuts about 1 second off the total compile time.


odisseus:
Looks like the layer is incorrect:
```
 53.996 [E] fury/strings>local:src/strings/diff.scala:27:41
 53.996 | not found: type ReadOnlyCaseClass
 53.996 | def combine[T: StringShow](caseClass: ReadOnlyCaseClass[Diff, T]): Diff[T] =
 54.048 [E] fury/strings>local:src/strings/diff.scala:117:29
 54.048 | not found: type ReadOnlyCaseClass
 54.048 | def combine[T](caseClass: ReadOnlyCaseClass[EntityName, T]): EntityName[T] =
```

propensive:
Yes, everything should have been updated in two versions of the build. I'm going to avoid that by updating the bootstrap version.

PR #1405, opened 2020-05-19
Reinstated tests

propensive:
Also renamed `strings` to `text`


propensive:
@odisseus I followed the steps you listed in PR #1301, though I had to do all the layer changes again due to the new format. I also made the tests run in parallel (not sure if it makes a difference in CI) and they are displayed nicely as a tree of suites at the end.

PR #1406, opened 2020-05-20
Updated the repo-pull test.

odisseus:


propensive:
It will certainly be nice not to see all the stack traces during the integration tests.

odisseus:
A more important change is the fact that this test should be passing since the issue #1370 was fixed.

propensive:
I think that there may still be an issue with it, at least with the text or the hashes that are displayed in the log messages. I think it's just a superficial issue, though.

PR #1412, opened 2020-05-20
node.js

propensive:
@odisseus I'm converting this branch into a PR so that we can track it.

### DIFF
--- a/src/build/build.scala
+++ b/src/build/build.scala
@@ -471,7 +471,7 @@ case class BuildCli(cli: Cli)(implicit log: Log) {
 
     compilation  <- Compilation.syncCompilation(schema, module.ref(project), layout,
                         https, false)
-    
+
     _            <- ~UiGraph.draw(compilation.graph.links, true,
                         Map())(ManagedConfig().theme).foreach(log.info(_))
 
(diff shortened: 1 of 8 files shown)


propensive:
@odisseus Given that we now have support for Scala JS, could we talk about the design ideas for this?

odisseus:
I created this branch because I was exploring the feasibility of using Fury as a dependency manager and build server for Node.js projects. Since there is no Scala code involved, I was going to bypass Bloop completely. Therefore, I don't see how this task is affected by the fact that Bloop now supports compiling Scala code to Javascript.

_P.S._ I just realized that now it's possible to compile whole Scala projects to Javascript and package them as huge fat Node packages. This is not what I aimed for, but it would be a major step towards full support of Node projects. (See #1440)

odisseus:
As for building Node.js packages with Fury, I think this is achievable, but it requires more work than I initially thought, especially since there are three or four different package formats that are used by the Node community.

propensive:
There wouldn't be any direct relationship between Scala.JS and Fury-as-a-JS-dependency-manager, except that when you're using Fury to produce some Scala.JS JavaScript, you would like to use the same tool to gather together your other JS dependencies.

veyga:
Would fury being acting as a replacement for npm, for a bundler (parcel/webpack), or as some sort of transpiler?

odisseus:
My original goal was building ready-to-use packages (which, I guess, is what Webpack does) that could use NPM packages as dependencies. The NPM dependencies would work just like binary dependencies in Scala projects. Of course, it would be possible to use source dependencies as an alternative.

PR #1418, opened 2020-05-21
Use ASM to get classname suggestions (which contain `main` methods)

### LINKED ISSUE #1417: provide tab-completions for class names
We would like it to be possible for users to press `<tab>` after a `--main` parameter, and have it suggest any classname that contains a `main` method with the appropriate signature.

propensive:


propensive:
@odisseus You may find this useful.

PR #1432, opened 2020-05-26
Preliminary support for Scala.js

propensive:


test/passing/scalajs/file.js
```diff
@@ -0,0 +1,1842 @@
+(function(){
```
odisseus, line 1: This looks like compilation output. Does this file indeed have to be stored in the repository?

PR #1441, opened 2020-05-29
Save Scala.js build artifacts as Node.js packages

### LINKED ISSUE #1440: Save Scala.js build artifacts as Node.js packages
See https://x-team.com/blog/a-guide-to-creating-a-nodejs-command/

Related issues are #1380 and #1431.

odisseus:
Closes #1440.

### DIFF
--- a/src/compile/compile.scala
+++ b/src/compile/compile.scala
@@ -589,12 +589,39 @@ case class Build(target: Target,
       }
     }
 
+    def packageJson(bin: Option[String]): euphemism.Json = {
+      //TODO move this to fury.utils.ScalaJs
+      import euphemism._
+      Json.of(
+        name = s"${ref.projectId.key}-${ref.moduleId.key}",
+        version = "0.0.1",
+        bin = bin
+      )
+    }
+
     def saveJs(staging: Path, module: Module): Try[Unit] = {
+      val jsName = str"${ref.projectId.key}-${ref.moduleId.key}.js"
       val dest = destination.extant()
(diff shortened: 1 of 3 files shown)


src/compile/compile.scala
```diff
@@ -570,30 +570,65 @@ case class Build(target: Target,
...
+    }
+
+    def packageJson(bin: Option[String]): euphemism.Json = {
+      //TODO move this to fury.utils.ScalaJs
+      import euphemism._
+      Json.of(
+        name = s"${ref.projectId.key}-${ref.moduleId.key}",
+        version = "0.0.1",
+        bin = bin
+      )
```
propensive, line 599: I think you can use a case class for this.

propensive (approved):
Looks great. I'm planning a big overhaul of the way different types of file are exported, but it will mainly affect the UI rather than this.

propensive:
By the way, I've left this unmerged in case you want to address the `TODO` and my comment first.

src/compile/compile.scala
odisseus: The Node.js package metadata defines [a lot](https://docs.npmjs.com/files/package.json) of fields, and I don't know yet which ones we will have to use. Someday these metadata are going to be wrapped in a case class, but I think it's too early at the moment.

odisseus:
The todo will be addressed in another pull request, because it requires modifying the layer file. The module `utils` doesn't define a dependency on Euphemism.

Issue #1396, opened 2020-05-19
`project list` should show inherited projects by default


propensive:
If we can show while layers the projects come from too, that would be good.

propensive:
And conflicts, if they exist.

odisseus:
What is the conceptual difference between `fury project list` and `fury layer projects`?

propensive:
`fury layer projects` was meant to show the entire universe, i.e. what I'm proposing here. I never really liked the confusion of these two commands though. I think `fury layer projects` could be removed.

odisseus:
The interface for listing projects could look like this:
* `fury project list` for just the current layer;
* `fury project list --tree` for the current layer and all imported layers, with an additional column for the layer name.

I'm not sure what exactly you mean by conflicts, but I believe that a layer with conflicting projects should be impossible to create in the first place.

PR #1451, opened 2020-06-10
List layers in projects view

propensive:


odisseus:
That's quite a lot of changes. Are all of them actually related to the issue #1396?

propensive:
Indirectly, but the original issue, as described in #1396, turned out not to be the solution I wanted. What I'm doing right now is more experimental.

Apart from the layer issue, does it conflict with anything you're doing?

odisseus:
At the moment I'm not doing anything complicated, so I don't think there will be significant conflicts.

However, I think that some of the changes in this pull request (at least the renaming of arguments) can be extracted into separate pull requests.

test/passing/repo-change-branch/script
```diff
@@ -0,0 +1,6 @@
+fury clean repositories
```
odisseus, line 1: The tests that use a pre-defined layer don't reduce complexity. They just hide it away in a format that's difficult to inspect. Additionally, these layers may become invalid as the layer format evolves.

test/passing/repo-change-branch/description
```diff
@@ -0,0 +1 @@
+Change to a different branch of a repository
```
odisseus, line 1: This test is missing a pre-defined layer.

test/passing/repo-branch-not-found/description
```diff
@@ -0,0 +1 @@
+Attempt to checkout a branch which does not exist.
```
odisseus, line 1: Looks like this test is missing a pre-defined layer.

test/passing/repo-change-tag/description
```diff
@@ -0,0 +1 @@
+Change to a different tag of a repository
```
odisseus, line 1: This test is missing a pre-defined layer.

PR #1460, opened 2020-07-14
`fury universe imports pull`

propensive:


src/core/resolver.scala
```diff
@@ -26,6 +26,7 @@ object Resolver {
   implicit val project: Resolver[Project, ProjectId] = _ == _.id
   implicit val repo: Resolver[Repo, RepoId] = _ == _.id
   implicit val imports: Resolver[Import, ImportId] = _ == _.id
+  implicit val importByShortRef: Resolver[Import, ShortLayerRef] = _.key == _.layerRef.key.drop(2).take(8)
```
odisseus: Consider using a pattern match with a regex:
```scala
case r"Qm$short@\w{6}\w{50}" => short
```

src/core/args.scala
```diff
@@ -76,8 +76,8 @@ object Args {
       "specify if this dependency should not be included transitively")
 
   val KeyArg = CliParam[String]('k', 'key, "GPG key")
-  val ImportArg = CliParam[LayerName]('i', 'layer, "specify an external layer to import")
-  val ImportIdArg = CliParam[ImportId]('l', 'layer, "specify a layer to unimport")
+  val ImportArg = CliParam[LayerName]('l', 'layer, "specify an external layer to import")
+  val ImportIdArg = CliParam[ImportId]('i', 'import, "specify an import by ID")
```
odisseus, line 80: Why?

src/core/resolver.scala
propensive: Yes, good point: the `Qm` is significant.

propensive: Though even better, I can just use the `.short` method on `LayerRef` and do an equality check.

src/core/args.scala
propensive: I think this might not be the end of this, actually. This is more like a snapshot on the way to working out the "best" names for these parameters. We have (I think) _four_ different ways of specifying a layer now:
- a remote name to import
- an Import ID in the current layer
- a layer path for making changes
- the short layer hash when selecting a layer from the universe

(There is also `-n` which is used to specify the new name of a layer, but I consider that distinct because it's a definition rather than a reference.)

I'd like to use as few different letters as possible for referencing these, with preference given to `-l` for the most common cases. But we can only have certain combinations of these appear on the same command, and they must all have distinct parameter names.

I'm not sure this fixes it, yet, and we may need to use three different letters due to the various combinations we need.

PR #1463, opened 2020-07-15
More manageable project conflicts

### LINKED ISSUE #1472: hash projects by repo commit
Currently, the hash used for projects is dependent only on the name of the repo, not on the resolved checkout. The `Universe` should have all the necessary information to resolve this.

propensive:
This changes the structure of the projects in a `Universe` to allow access to a `Map` of projects by `ProjectId`, or an exact project by `ProjectRef`. If there is a unique definition for a project, the inner `Map` will have just a single entry, whereas conflicting project definitions will be represented by a map with multiple entries, from `ProjectRef` to `Set[ImportPath]`, which references the layers containing those projects.

This facilitates the `fury universe projects list` command which will list all the projects in the universe, including conflicts. However conflicting projects will be displayed, and must be referenced to by their project ID _and_ a six-character hex hash, something like, `magnolia#17fba9`.

This table of conflicting projects is also displayed when attempting to build while conflicts exist. For now, this just introduces the list of projects, though there will be some more PRs with changes that make it possible to manipulate the universe better.

### DIFF
--- a/src/core/universe.scala
+++ b/src/core/universe.scala
@@ -33,21 +33,43 @@ case class Universe(hierarchy: Hierarchy,
                     imports: Map[ShortLayerRef, LayerEntity]) {
   def ids: Set[ProjectId] = projects.keySet
   
+  def importPaths(id: ProjectRef): Try[Set[ImportPath]] = projects.get(id.id).ascribe(ItemNotFound(id)).flatMap {
+    case map if map.size == 1 =>
+      Success(map.values.head)
+    case map                  =>
+      map.get(id).ascribe(ProjectConflict(map.to[List].traverse { case (ref, layers) =>
+        apply(ref).map((ref, _, layers))
+      }.getOrElse(Nil)))
+  }
+
   def importPaths(id: ProjectId): Try[Set[ImportPath]] = projects.get(id).ascribe(ItemNotFound(id)).flatMap {
-    case map if map.size == 1 => Success(map.values.head)
-    case map                  => Failure(ProjectConflict(map))
+    case map if map.size == 1 =>
+      Success(map.values.head)
+    case map =>
+      Failure { ProjectConflict { map.to[List].traverse { case (ref, layers) =>
+        apply(ref).map((ref, _, layers))
+      }.getOrElse(Nil) } }
   }
 
   def allProjects: Try[Set[Project]] = projects.keySet.traverse(apply(_))
   def layer(id: ProjectId): Try[Layer] = importPaths(id).flatMap { is => hierarchy(is.head) }
-  def layer(id: ProjectRef): Try[Layer] = hierarchy(projects(id.id)(id).head)
+  
+  def layer(id: ProjectRef): Try[Layer] = (projects(id.id) match {
+    case map if map.size == 1 =>
+      Try(map.head._2.head)
+    case map =>
+      map.get(id).orElse(map.find(_._1.id == id.id).map(_._2)).map(_.head).ascribe(ItemNotFound(id))
+  }).flatMap(hierarchy(_))
+  
   def apply(id: ProjectRef): Try[Project] = layer(id).flatMap(_.projects.findBy(id.id))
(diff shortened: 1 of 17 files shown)


src/core/universe.scala
```diff
@@ -24,23 +24,61 @@ import gastronomy._
...
+  def importPaths(id: ProjectRef): Try[Set[ImportPath]] = projects.get(id.id).ascribe(ItemNotFound(id)).flatMap {
+    case map if map.size == 1 =>
+      Success(map.values.head)
+    case map                  =>
+      map.get(id).ascribe(ProjectConflict(map.to[List].traverse { case (ref, layers) =>
+        apply(ref).map((ref, _, layers))
+      }.getOrElse(Nil)))
+  }
+
+  def importPaths(id: ProjectId): Try[Set[ImportPath]] = projects.get(id).ascribe(ItemNotFound(id)).flatMap {
```
odisseus, line 45: This method and the previous one can be easily confused. It took me quite some time to realize what is the difference between them and why they need to be different.

src/core/universe.scala
```diff
@@ -24,23 +24,61 @@ import gastronomy._
 import scala.util._
 import scala.collection.immutable.TreeSet
 
-object Universe { def apply(): Universe = Universe(Map(), Map(), Map()) }
+object Universe { def apply(hierarchy: Hierarchy): Universe = Universe(hierarchy, Map(), Map(), Map()) }
 
 /** A Universe represents a the fully-resolved set of projects available in the layer */
-case class Universe(entities: Map[ProjectId, Entity],
+case class Universe(hierarchy: Hierarchy,
+                    projects: Map[ProjectId, Map[ProjectRef, Set[ImportPath]]],
```
odisseus, line 31: It would be nice to have domain classes for project references during the universe construction, e. g. like this:
```scala
sealed trait ProjectDefinition { 
  def +(other: ProjectDefinition): ProjectDefinition 
}
case class UniqueProject(ref: ProjectRef, origins: Set[ImportPath]) extends ProjectDefinition
case class ConflictingProjects private(origins: Map[ImportPath, ProjectRef]) extends ProjectDefinition
```

src/core/universe.scala
```diff
@@ -24,23 +24,61 @@ import gastronomy._
...
-  def ids: Set[ProjectId] = entities.keySet
-  def entity(id: ProjectId): Try[Entity] = entities.get(id).ascribe(ItemNotFound(id))
-  def spec(id: ProjectId): Try[ProjectSpec] = entity(id).map(_.spec)
+  def ids: Set[ProjectId] = projects.keySet
+  
+  def importPaths(id: ProjectRef): Try[Set[ImportPath]] = projects.get(id.id).ascribe(ItemNotFound(id)).flatMap {
+    case map if map.size == 1 =>
+      Success(map.values.head)
+    case map                  =>
+      map.get(id).ascribe(ProjectConflict(map.to[List].traverse { case (ref, layers) =>
```
odisseus, line 40: I'm not sure "ProjectConflict" is the correct name for the exception that might happen here. A project reference is always resolved in an unique way, and conflicts with other references shouldn't affect this.

This exception would happen if the `projects` map contained an entry for the project ID , but not for the project reference. I'd say this is an `IllegalStateException`, or maybe an `ItemNotFound`.

src/core/universe.scala
propensive: That's a fair point. The same could be said for the `apply` methods.

src/core/universe.scala
propensive: I agree. I actually had this same idea while over the weekend.

src/core/universe.scala
propensive: But those exceptions don't give us a way to pass the useful information about the unresolved projects back to the user... but could we call it `AmbiguousProjects` instead, maybe?

odisseus: If I understand your intention correctly, this exception just shouldn't happen. If it ever happens, that would mean a bug in Fury itself rather than a problem with the layer definition, and a stack trace would be more appropriate than a user-friendly error.

PR #1464, opened 2020-07-15
Fixed resolution of the local checkout directory for repositories.

### LINKED ISSUE #1462: Bloop configuration is broken for forks of local repositories
This is the reason why the repo-fork test fails.

**Before fork:**
```
"sources": [
    "/home/user/.cache/fury/sources/C25973C9AED0F365/src/core"
],
```

**After fork:**
```
"sources": [
    "/home/user/devel/fury/test/tmp/repo-fork/app/home/user/devel/fury/test/tmp/repo-fork/app/library-local/src/core"
],
```

This issue was caused by PR #1451.

odisseus:
Fixes #1462.

### DIFF
--- a/src/core/sourceRepo.scala
+++ b/src/core/sourceRepo.scala
@@ -63,7 +63,7 @@ case class Repo(id: RepoId, remote: Remote, branch: Branch, commit: Commit, loca
     Snapshot(id, remote, localDir(layout), commit, branch, List())
 
   def localDir(layout: Layout)(implicit log: Log): Option[GitDir] =
-    local.map { p => GitDir(p in layout.baseDir)(layout.env) }
+    local.map { p => GitDir(layout.baseDir.resolve(p))(layout.env) }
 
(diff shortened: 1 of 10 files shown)


src/core/sourceRepo.scala
```diff
@@ -63,7 +63,7 @@ case class Repo(id: RepoId, remote: Remote, branch: Branch, commit: Commit, loca
     Snapshot(id, remote, localDir(layout), commit, branch, List())
 
   def localDir(layout: Layout)(implicit log: Log): Option[GitDir] =
-    local.map { p => GitDir(p in layout.baseDir)(layout.env) }
```
propensive, line 66: I wonder if we should rewrite `in`. Is what it's currently doing ever useful?

propensive (approved):

src/core/sourceRepo.scala
odisseus: I'm not sure. Currently, `a in b` is equivalent to `b / a.value`. However, the method `/` accepts strings which may contain something like `..` and don't even have to be valid paths.

PR #1475, opened 2020-07-20
Domain classes for project references inside a Universe

odisseus:
I have tried to keep the logic unchanged. Please see my comments to the PR #1463.

### DIFF
--- a/src/core/universe.scala
+++ b/src/core/universe.scala
@@ -24,41 +24,77 @@ import gastronomy._
 import scala.util._
 import scala.collection.immutable.TreeSet
 
-object Universe { def apply(hierarchy: Hierarchy): Universe = Universe(hierarchy, Map(), Map(), Map()) }
+object Universe {
+  def apply(hierarchy: Hierarchy): Universe = Universe(hierarchy, Map(), Map(), Map())
+
+  sealed trait ProjectDefinition {
+    def id: ProjectId
+    def +(other: ProjectDefinition): ProjectDefinition
+  }
+
+  case class UniqueProject(ref: ProjectRef, origins: Set[Pointer]) extends ProjectDefinition {
+    override def id: ProjectId = ref.id
+
+    override def +(other: ProjectDefinition): ProjectDefinition = {
+      if(other.id != this.id) throw new IllegalArgumentException(str"Project IDs ${this.id} and ${other.id} are different")
+      else other match {
+        case UniqueProject(ref, origins) if ref == this.ref => UniqueProject(ref, this.origins ++ origins)
+        case ConflictingProjects(origins) => ConflictingProjects(origins ++ this.origins.map(i => i -> this.ref))
+      }
+    }
+  }
+
+  case class ConflictingProjects private[Universe](origins: Map[Pointer, ProjectRef]) extends ProjectDefinition {
+    override def id: ProjectId = origins.values.head.id
+
+    override def +(other: ProjectDefinition): ProjectDefinition = {
+      if (other.id != this.id) throw new IllegalArgumentException(str"Project IDs ${this.id} and ${other.id} are different")
+      else other match {
+        case UniqueProject(ref, origins) => ConflictingProjects(this.origins ++ origins.map(i => i -> ref))
+        case ConflictingProjects(origins) => ConflictingProjects(this.origins ++ origins)
+      }
+    }
+  }
+}
 
 /** A Universe represents a the fully-resolved set of projects available in the layer */
 case class Universe(hierarchy: Hierarchy,
-                    projects: Map[ProjectId, Map[ProjectRef, Set[Pointer]]],
+                    projects: Map[ProjectId, Universe.ProjectDefinition],
                     repoSets: Map[RepoSetId, Set[RepoRef]],
                     imports: Map[ShortLayerRef, LayerEntity]) {
   def ids: Set[ProjectId] = projects.keySet
-  
+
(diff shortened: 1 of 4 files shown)


src/repo/universe.scala
```diff
@@ -126,7 +126,10 @@ case class UniverseApi(hierarchy: Hierarchy) {
       layer        <- hierarchy(importPath)
       universe     <- hierarchy.universe
       project      <- universe(projectRef)
-      importPaths  <- ~universe.projects(projectRef.id)(projectRef)
+      importPaths  <- ~(universe.projects(projectRef.id) match {
+        case Universe.UniqueProject(_, origins) => origins
+        case Universe.ConflictingProjects(origins) => origins.keys
+      })
```
propensive: Can we move this logic into a method on `Universe`?

src/core/universe.scala
```diff
@@ -24,41 +24,77 @@ import gastronomy._
 import scala.util._
 import scala.collection.immutable.TreeSet
 
-object Universe { def apply(hierarchy: Hierarchy): Universe = Universe(hierarchy, Map(), Map(), Map()) }
+object Universe {
+  def apply(hierarchy: Hierarchy): Universe = Universe(hierarchy, Map(), Map(), Map())
+
+  sealed trait ProjectDefinition {
```
propensive: I the interests of concision, maybe we could call this `ProjectDef`.

propensive: And I think it could be a top-level trait. There is an argument for making it an inner trait of the `Universe` class, on the basis that uniqueness or ambiguity are dependent on the universe, but I'm not sure that's worth the effort.

src/core/universe.scala
```diff
@@ -24,41 +24,77 @@ import gastronomy._
...
+    def id: ProjectId
+    def +(other: ProjectDefinition): ProjectDefinition
+  }
+
+  case class UniqueProject(ref: ProjectRef, origins: Set[ImportPath]) extends ProjectDefinition {
```
propensive: Let's move this inside the `ProjectDef` companion object, and call it `Unique`.

src/core/universe.scala
```diff
@@ -24,41 +24,77 @@ import gastronomy._
...
+    override def +(other: ProjectDefinition): ProjectDefinition = {
+      if(other.id != this.id) throw new IllegalArgumentException(str"Project IDs ${this.id} and ${other.id} are different")
+      else other match {
+        case UniqueProject(ref, origins) if ref == this.ref => UniqueProject(ref, this.origins ++ origins)
+        case ConflictingProjects(origins) => ConflictingProjects(origins ++ this.origins.map(i => i -> this.ref))
+      }
+    }
+  }
+
+  case class ConflictingProjects private[Universe](origins: Map[ImportPath, ProjectRef]) extends ProjectDefinition {
```
propensive: And let's move this into `ProjectDef` and call it `Ambiguous`, so it would be `ProjectDef.Ambiguous`.

src/core/universe.scala
```diff
@@ -24,41 +24,77 @@ import gastronomy._
...
+  sealed trait ProjectDefinition {
+    def id: ProjectId
+    def +(other: ProjectDefinition): ProjectDefinition
+  }
+
+  case class UniqueProject(ref: ProjectRef, origins: Set[ImportPath]) extends ProjectDefinition {
+    override def id: ProjectId = ref.id
+
+    override def +(other: ProjectDefinition): ProjectDefinition = {
+      if(other.id != this.id) throw new IllegalArgumentException(str"Project IDs ${this.id} and ${other.id} are different")
```
propensive: I realize that this exception, if thrown, represents a programming error, though I'm not sure we want to actually throw anything. Could we move the logic away from having `+` be a member of `ProjectDef` and to some place where `other.id == this.id` is already asserted?

propensive (changes requested):
One other possibility may be to create a sealed generic type, `Uniqueness[T]` with subtypes, `Unique[T]` and `Ambiguous[T]`, rather than having a `Project`-specific one. I think it would be generally useful, and you wouldn't be able to rely on any particular features of `Project` in its definition.

src/core/universe.scala
odisseus: The difference between unique and conflicting project definitions makes sense only during the universe construction, because the command execution just wouldn't proceed beyond that stage if there are conflicts. To emphasise that fact and to reduce clutter in the core namespace, I have decided to put these classes inside the `Universe` object, even thought they aren't really dependent on any particular universe.

src/repo/universe.scala
odisseus: Done

src/core/universe.scala
odisseus: Moved into a separate file.

I have moved the trait to the top level, but the case classes need to be inside an object (or a small dedicated package) because I want to hide the constructor for the `Ambiguous` class. It should be possible to create it only by merging `Unique` objects that aren't compatible with each other.

src/core/universe.scala
odisseus: Renamed

src/core/universe.scala
odisseus: Renamed

src/core/universe.scala
odisseus: I have removed this exception altogether, because this operator is used only once, and the context makes an ID mixup quite unlikely.

PR #1485, opened 2020-07-27
Integration tests for compiler plugins

odisseus:
I had to make the unapply regex for `OptId` more permissive, because a plugin module defines an option that looks like this:
```
┌───┬─────────────────────────────────────────────────────┬────────────┬────────┐
│   │ Param                                               │ Persistent │ Origin │
├───┼─────────────────────────────────────────────────────┼────────────┼────────┤
│ + │ Xplugin:/base/dir/.fury/classes/project-name_module │ Yes        │ plugin │
└───┴─────────────────────────────────────────────────────┴────────────┴────────┘

```
This option is inherited by all modules that depend on the plugin, and it could be neither removed nor added back.

### DIFF
--- a/test/passing/plugin-jar/src/app.scala
+++ b/test/passing/plugin-jar/src/app.scala
@@ -1,0 +1,22 @@
+package app
+
+object Main {
+
+  // Copied from https://github.com/typelevel/kind-projector/blob/master/src/test/scala/bounds.scala
+  trait Leibniz[-L, +H >: L, A >: L <: H, B >: L <: H]
+
+  object Test {
+    trait Foo
+    trait Bar extends Foo
+
+    def outer[A >: Bar <: Foo] = {
+      def test[F[_ >: Bar <: Foo]] = "Hello kind-projector!"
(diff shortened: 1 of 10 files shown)


test/passing/plugin-jar/src/app.scala
```diff
@@ -0,0 +1,22 @@
...
+  // Copied from https://github.com/typelevel/kind-projector/blob/master/src/test/scala/bounds.scala
+  trait Leibniz[-L, +H >: L, A >: L <: H, B >: L <: H]
+
+  object Test {
+    trait Foo
+    trait Bar extends Foo
+
+    def outer[A >: Bar <: Foo] = {
+      def test[F[_ >: Bar <: Foo]] = "Hello kind-projector!"
+      test[λ[`b >: Bar <: Foo` => Leibniz[Bar, Foo, A, b]]]
```
propensive, line 14: Unicode test for free!

propensive (approved):
All looks good.

odisseus:
I just noticed that both builds on Mac OS have failed with differences in the test output. Maybe it was too early to merge this...

https://github.com/propensive/fury/runs/917352511?check_suite_focus=true#step:9:64

PR #1488, opened 2020-07-28
Fix bad root/empty pointers

### LINKED ISSUE #1487: importing and cloning fail for layers with dependent layers

propensive:
This PR fixes a problem that became apparent when cloning or importing a layer that module references couldn't be resolved. Somewhere, we were appending an import ID to a root path, with a `/` between them, which resulted in a pointer beginning with `//`, which then failed to match.

This changes the definition of `def /` so that it first checks if it's appending to an empty layer, and omits the `/` in those cases.

I also provided a more convenient way of logging to the global log (without a PID) using `Log().info` or `Log().note` etc. I've also added the `--ignore`/`-I` flag to `layer clone` and `layer import` which permits imports even when they're failing, though this should only be useful temporarily for working around bugs like this, because publishing of broken layers is still impossible.

### DIFF
--- a/src/core/args.scala
+++ b/src/core/args.scala
@@ -75,6 +75,7 @@ object Args {
   val IntransitiveArg = CliParam[Unit]('I', 'intransitive,
       "specify if this dependency should not be included transitively")
 
+  val IgnoreArg = CliParam[Unit]('I', 'ignore, "ignore layer errors")
   val KeyArg = CliParam[String]('k', 'key, "GPG key")
   val ImportArg = CliParam[LayerName]('l', 'layer, "specify an external layer to import")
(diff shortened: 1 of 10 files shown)


src/core/args.scala
```diff
@@ -75,6 +75,7 @@ object Args {
   val IntransitiveArg = CliParam[Unit]('I', 'intransitive,
       "specify if this dependency should not be included transitively")
 
+  val IgnoreArg = CliParam[Unit]('I', 'ignore, "ignore layer errors")
```
odisseus, line 78: I think it would be better to reuse `ForceArg`. It would be harder to confuse it with `ImportIdArg`.

PR #1489, opened 2020-07-31
Refactored the commands related to modules to use the new CLI API

odisseus:


src/module/module.scala
```diff
@@ -44,7 +45,19 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
                      yield Plugin(p, m)
   }
-  
+
+  private def resolveToCompiler(ref: CompilerRef): Try[CompilerRef] = {
+    ref match {
+      case Javac(n) => ~(ref)
+      case bsp@BspCompiler(ref) =>
+        for {
+          available <- (getLayer, getLayout) >> (_.compilerRefs(_))
```
propensive: This could go into the `CliApi` as a reusable command.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
-
-    layer       <- if(newId.isEmpty || project.main != Some(module.id)) ~layer
-                   else ~(Layer(_.projects(project.id).main)(layer) = newId)
-
-    layer       <- ~name.fold(layer)(Layer(_.projects(project.id).modules(module.id).id)(layer) = _)
-    _           <- Layer.commit(layer, conf, layout)
-    _           <- ~Build.asyncBuild(layer, module.ref(project), layout)
-  } yield log.await()
+  def select: Try[ExitStatus] = {
+    (cli -< ProjectArg -< ModuleArg).action {
```
propensive: I know the original version didn't have it, but we should also take `LayerArg` now.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+      (getProject, cliModule >> (_.id) >> (Option(_))) >>= { case (proj, optMain) =>
+        (getLayer, mainModuleLens) >> Layer.set(optMain) >> commit >> finish
+      }
+    }
+  }
+
+  def list: Try[ExitStatus] = {
+    implicit val columns: ColumnArg.Hinter =
+      ColumnArg.hint("module", "dependencies", "sources", "binaries", "compiler", "options", "type")
```
propensive: We should be able to get the list of columns from the `table` object, if we define it (lazily) outside of the `action` block.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+  def add: Try[ExitStatus] = {
+    implicit val dummy: ModuleNameArg.Hinter = ModuleNameArg.hint()
+    implicit val dummy2: HiddenArg.Hinter = HiddenArg.hint()
+    implicit val dummy3: CompilerArg.Hinter = CompilerArg.hint()
+    implicit val dummy4: MainArg.Hinter = MainArg.hint()
+    implicit val dummy5: ReplArg.Hinter = ReplArg.hint()
+    implicit val dummy6: TimeoutArg.Hinter = TimeoutArg.hint()
+    implicit val dummy7: SpecArg.Hinter = SpecArg.hint()
+    implicit val dummy8: PluginArg.Hinter = PluginArg.hint()
+    implicit val foo: StringShow[Boolean] = { b => if (b) "on" else "off" }
```
propensive: I think we should always try to avoid names like `dummy`, though I understand it's a reasonable placeholder, just to get things compiling. Obviously these all need actual implementations... I also forgot there were so many different parameters involved.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+      MainArg -< ReplArg -< TimeoutArg -< SpecArg -< PluginArg).action {
+      val newModule = for {
+        newModuleId <- get(ModuleNameArg)
+        _ <- getProject >>= (_.modules.unique(newModuleId))
+        kind = cliKind.getOrElse(Lib())
+        compiler <- cliCompiler.orElse(defaultCompiler)
+        hidden = has(HiddenArg)
+      } yield {
+        Module(newModuleId, compiler = compiler, kind = kind, hidden = hidden)
+      }
```
propensive: That should be a single line.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+      for {
+        _ <- newLayer >>= commit
+        _ <- (newLayer, (newModule, getProject) >> (_.ref(_)), getLayout) >> Build.asyncBuild
+      } yield log.await()
+    }
+  }
+
+  def remove: Try[ExitStatus] = {
+    (cli -< ProjectArg -< ModuleArg).action {
```
propensive, line 83: This should also have `LayerArg`.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+      val newLayer = for {
+        newLayer <- (newModules, getLayer, modulesLens) >> (Layer.set(_)(_, _))
+        cleanMain <- (getProject >> (_.main), cliModule) >> (_.contains(_))
+        lens <- mainModuleLens
+      } yield {
+        if(cleanMain) Layer.set(Option.empty[ModuleId])(newLayer, lens)
+        else newLayer
+      }
+      for {
+        _ <- newLayer >>= commit
```
propensive, line 94: There's a command (somewhere... you'll have to check) for updating any layer within the hierarchy in a single command. The `RepoApi` will have it, I think.

src/module/module.scala
```diff
@@ -66,158 +79,199 @@ case class ModuleCli(cli: Cli)(implicit log: Log) {
...
+
+      layer       <- if(newId.isEmpty || project.main != Some(module.id)) ~layer
+      else ~(Layer(_.projects(project.id).main)(layer) = newId)
+
+      layer       <- ~name.fold(layer)(Layer(_.projects(project.id).modules(module.id).id)(layer) = _)
+      _           <- Layer.commit(layer, conf, layout)
+      _           <- ~Build.asyncBuild(layer, module.ref(project), layout)
+    } yield log.await()
+  }
+
```
propensive: I forgot how huge this had become...

It might be possible to reuse significant parts of this and the `add` implementation.

propensive (commented):
Looks like good progress. I'm keen on the idea of iterating on the "components" we join together here once we've got something working.

One possible improvement to the approach may be to give more components identifiers (rather than inlining them) so that each command can finish up with a single line which reads something like (just a meaningless example),
```
getModule >>= setCompiler >>= setName >>= setSomethingElse >>= finish
```
to serve as a very high-level view of what is happening. But there are hundreds of different ways we could format this, and I would expect a lot of experimentation before finding the clearest way.

No need to let this experimentation hold up what you're doing already, though.

src/module/module.scala
odisseus: Refactored into a private lazy val. It can be moved to `CliApi` later if the other command sets need it.

src/module/module.scala
odisseus: Okay, but I'm not sure that's actually an improvement. In this case, the column headings are known at compile time, and they could be hinted even when there's a problem with resolving the project or constructing the universe.

src/module/module.scala
odisseus: Ok

src/module/module.scala
propensive: But for maintainability, we don't have those columns listed in two different places which can get out of sync. The `table` value should not depend on any data, so it ought to be possible to get the column headings.

src/module/module.scala
odisseus: It doesn't seem to work, not even for the repo commands.
```
$ fury repo list --layer java-ipfs-http-client
┌──────┬────────┬────────────┬────────┬──────┬─────────┐
│ Repo │ Remote │ Branch/Tag │ Commit │ Path │ Changes │
├──────┼────────┼────────────┼────────┼──────┼─────────┤
└──────┴────────┴────────────┴────────┴──────┴─────────┘
$ fury layer select --layer java-ipfs-http-client
$ fury repo list
┌───────────────────────┬───────────────────────────────────┬──────────────────┬─────────┬──────┬─────────┐
│ Repo                  │ Remote                            │ Branch/Tag       │ Commit  │ Path │ Changes │
├───────────────────────┼───────────────────────────────────┼──────────────────┼─────────┼──────┼─────────┤
│ java-cid              │ gh:odisseus/java-cid              │ multihash-update │ 8398ca3 │ -    │ -       │
│ java-ipfs-http-client │ gh:odisseus/java-ipfs-http-client │ post-go-ipfs-0.5 │ f395089 │ -    │ -       │
│ java-multiaddr        │ gh:multiformats/java-multiaddr    │ master           │ 2347e9f │ -    │ -       │
│ java-multibase        │ gh:odisseus/java-multibase        │ base32-fix       │ 51c2a0b │ -    │ -       │
│ java-multihash        │ gh:multiformats/java-multihash    │ master           │ 0eb305c │ -    │ -       │
└───────────────────────┴───────────────────────────────────┴──────────────────┴─────────┴──────┴─────────┘

```

src/module/module.scala
odisseus: You probably mean [`fury.core.Hierarchy#update`](https://github.com/propensive/fury/blob/master/src/core/hierarchy.scala#L41), but using it doesn't make sense as long as the way to select any layer doesn't work.

src/module/module.scala
odisseus: I have removed all of the dummy hinters, but the hints are still lacking when compared to the master version.

PR #1496, opened 2020-08-04
Introduce `includes` submenu

propensive:
We can now list, add, remove and update basic exports from dependencies into the workspace of modules using commands like,
```
fury include add -f classdir -t classes -d escritore/core
fury include list
fury include update -e escritoire-classes -n esc
fury include remove -e esc
```


src/source/source.scala
```diff
@@ -16,7 +16,7 @@
 */
 package fury
 
-import fury.text._, fury.core._, fury.model._
+import fury.text._, fury.core._, fury.model._, fury.io._
```
odisseus, line 19: Why?

propensive: I'm pretty sure it was added because `Path` was used in the `ExportCli` menu, which I started working on in the `src/source/source.scala` file, before I moved to its own file, and this was left over afterwards.

propensive:
I think `hello-world` is failing because the PWD of `app` modules has changed, prohibiting writing to the layer root directory.

PR #1486, opened 2020-07-28
Replaced various equivalent regular expressions with [\w\-\.]

odisseus:

PR #1504, opened 2020-08-09
Improve binary handling

propensive:
The handling of binaries has been improved. We now use Coursier to resolve all the binaries for a particular module together, rather than one-by-one. This may result in Coursier making decisions about automatically upgrading certain transitive dependencies that it would not have otherwise done, but Fury does not make any judgement on those choices.

When the binary dependencies of modules get combined (because a module defines a module dependency and binary dependencies, or, a module defines more than one dependency) then the fully resolved set of dependencies will be compared to check whether there exist any different versions of the same group/artifact combination in the combined tree, and flags this as a conflict, refusing to build. The conflicts are listed in the error message.

The conflict can usually be resolved by adding specific binaries to one or other of the modules.

This approach ensures a useful invariant: that a module will use exactly the same set of binary dependencies whether it is built alone, or whether it is built as a dependency.

A few more changes to improve the error message and fix some tests are required before this is ready for merge.

This PR also includes changes to the layer to remove conflicts in the resolution of binaries.


src/compile/compile.scala
```diff
@@ -482,21 +482,24 @@ case class Build(target: Target,
...
     include.kind match {
-      case IncludeType.ClassesDir =>
+      case ClassesDir(deep) =>
         include.path.mkdir().flatMap {
           layout.classesDir(include.ref).copyTo(include.path in layout.workDir(ref)).waive
         }.munit
-      case IncludeType.FileRef(glob) =>
-        glob(layout.workDir(include.ref), layout.workDir(include.ref).walkTree).traverse { p =>
+      case FileRef(source) =>
+        /*glob(layout.workDir(include.ref), layout.workDir(include.ref).walkTree).traverse { p =>
```
odisseus, line 490: Why?

src/core/cli.scala
```diff
@@ -361,6 +361,8 @@ abstract class CliApi {
     case other                        => other.map(Some(_))
   }
 
+  def boolean(arg: CliParam): Try[Boolean] = opt(arg).map(_.isDefined)
```
odisseus, line 365: How is this going to be using in practice? Do I understand correctly that this method should return `Success(true)` e.g. when the user types `--gzip on` and `Success(false)` when the user types `--gzip off`?

src/core/cli.scala
```diff
@@ -499,6 +522,25 @@ abstract class CliApi {
...
   def printTable[T, S: MsgShow](table: Tabulation[T], rows: Traversable[T], id: Option[S], name: String): Unit =
     log.rawln(Tables().show(table, cols, rows, raw, column, id, name))
+  
+  private[this] def possibleSourceDirectories(repo: Repo, layout: Layout) =
+    repo.sourceCandidates(layout)(isSourceFileName).getOrElse(Set.empty)
+  
+  private[this] def isSourceFileName(name: String): Boolean =
+    name.endsWith(".scala") || name.endsWith(".java")
+  
+  object sourceHints {
```
odisseus, line 555: Are these hinters useful anywhere outside `SourceCli`?

src/core/shell.scala
```diff
@@ -83,6 +83,12 @@ case class Shell(environment: Environment) {
     Success(())
   }
 
+  object docker {
+    def images(): Try[List[ImageId]] = sh"".exec[Try[String]].map { str =>
```
odisseus: The actual shell command is missing.

src/frontend/recovery.scala
```diff
@@ -58,6 +58,12 @@ object Recovery {
           cli.call()
           cli.layout.foreach(Layer.showMergeConflicts(_))
           cli.abort(msg"It is not possible to continue until merge conflicts have been resolved.")
+        case BinaryConflict(map) =>
+          val conflicts = map.to[List].map { case (name, versions) => (name, versions.mapValues(_.version)) }
+          val table = Tables().show(Tables().binaryConflicts, cli.cols, conflicts, false, None, None: Option[String], "binary")
+          log.info("This module contains conflicting binaries.")
```
odisseus, line 64: The error message would be more clear if it included the name of the affected module.

src/model/ids.scala
```diff
@@ -553,48 +553,18 @@ object Scope {
...
-    case "classes"          => ClassesDir
-    case r"file:$glob@(.+)" => FileRef(Glob(glob))
+object ImageId {
+  implicit val msgShow: MsgShow[ImageId] = m => UserMsg(_.image(m.key))
+  implicit val stringShow: StringShow[ImageId] = _.key
+  implicit val diff: Diff[ImageId] = (l, r) => Diff.stringDiff.diff(l.key, r.key)
+  implicit val parser: Parser[ImageId] = unapply(_)
+
+  def unapply(str: String): Option[ImageId] = str.only {
+    case r"[_a-zA-Z0-9][_.\-a-zA-Z0-9]*[_a-zA-Z0-9]" => ImageId(str)
```
odisseus, line 563: The list `[_a-zA-Z0-9]` is equivalent to `\w`. See #1486.

src/model/kinds.scala
```diff
@@ -29,36 +29,41 @@ object Kind {
...
+    case App(main, timeout, ws)        => msg"app${':'}$main"
+    case Container(image, timeout, ws) => msg"container${':'}$image"
+    case Plugin(id, main)              => msg"plugin${':'}$id${'@'}$main"
+    case Compiler(spec, repl)          => msg"compiler${':'}$spec"
+    case Bench(main)                   => msg"bench${':'}$main"
   }
 }
 
-sealed abstract class Kind(val needsExec: Boolean = false) {
+sealed abstract class Kind(val needsExec: Boolean = false) extends scala.Product with scala.Serializable {
```
odisseus, line 45: Why?

odisseus (changes requested):
Can the introduction of workspaces, Docker images etc. be separated from the changes to Coursier interaction?

By the way, the unit tests fail with [compilation errors](https://github.com/propensive/fury/runs/990675652?check_suite_focus=true#step:7:114):
```
 51.054 [E] fury/text-test>local:src/text-test/prefixes.scala:27:26
 51.054 | not found: type Suite
 51.054 | object TextTests extends Suite() {
 51.155 Compilation of module fury/text-test failed
```

src/compile/compile.scala
propensive: I can't remember... but it looks like something I'll need to sort out when I get back there...

src/core/cli.scala
propensive: It's intended for "present"/"not present" checking, for parameters that are parameterized on `Unit`. IIRC, it converts them from `Success(())` to `Success(true)` and `Failure(_)` to `Success(false)`.

src/core/cli.scala
propensive: Potentially... I think there are actually references to them from two different files, though I don't remember which.

src/core/shell.scala
propensive: Thanks - I think I was actually confused why I wasn't getting Docker image suggestions...

src/frontend/recovery.scala
propensive: There is _lots_ more information we could show here, including strategies for fixing the problem.

src/model/kinds.scala
propensive: This has the effect that the inferred LUB of two or more of `Kind`'s subtypes will be `Kind` rather than `Kind with Product with Serializable`. It makes some errors nicer.

src/model/ids.scala
propensive: I'll change it. I can never remember all these character classes...

PR #1508, opened 2020-08-12
Don't include globs into completion hints for source commands

### LINKED ISSUE #1505: source completions are wrong
Source completions do not contain any entries from external repos, and also complete full file paths instead of directories.

odisseus:
Fixes #1505.


propensive (approved):
You're right, it was easy!

PR #1512, opened 2020-08-14
Command to replace all entries of a layer import with another layer

### LINKED ISSUE #1510: The universe imports pull should allow to update any layer reference with any other layer reference

odisseus:
Closes #1510.

### DIFF
--- a/src/frontend/menu.scala
+++ b/src/frontend/menu.scala
@@ -167,7 +167,7 @@ object FuryMenu {
       ),
       Menu('imports, msg"view and modify layer imports", 'list, shortcut = 'i')(
         Action('list, msg"show all layer imports", UniverseCli(_).imports.list, shortcut = 'l'),
-        Action('pull, msg"update layer imports universally", UniverseCli(_).imports.pull, shortcut = 'p')
+        Action('update, msg"update layer imports universally", UniverseCli(_).imports.update, shortcut = 'u'),
       )
     ),
     Menu('layer, msg"view and edit the layer", 'projects, shortcut = 'l', needsLayer = false)(
--- a/src/repo/universe.scala
+++ b/src/repo/universe.scala
@@ -95,30 +95,34 @@ case class UniverseCli(cli: Cli)(implicit val log: Log) extends CliApi {
     lazy val table: Tabulation[LayerEntity] = Tables().layerRefs
     implicit val columnHints: ColumnArg.Hinter = ColumnArg.hint(table.headings.map(_.name.toLowerCase))
 
-    def list: Try[ExitStatus] = (cli -< RawArg -< ColumnArg).action { for {
-      col      <- opt(ColumnArg)
-      layerRef <- opt(LayerRefArg)
-      rows     <- universe >> (_.imports.values.to[List])
-      table    <- ~Tables().show(table, cli.cols, rows, has(RawArg), col, layerRef >> (_.key), "layer")
-      _        <- conf >> (_.focus()) >> (log.infoWhen(!has(RawArg))(_))
-      _        <- ~log.rawln(table)
-    } yield log.await() }
+    def list: Try[ExitStatus] = (cli -< RawArg -< ColumnArg).action {
+      val output = (opt(ColumnArg), opt(LayerRefArg), universe >> (_.imports.values.to[List])) >> { case (col, layerRef, rows) =>
+        Tables().show(table, cli.cols, rows, has(RawArg), col, layerRef >> (_.key), "layer")
+      }
+      for {
+        _ <- conf >> (_.focus()) >> (log.infoWhen(!has(RawArg))(_))
+        _ <- output >> log.rawln
+      } yield log.await()
+    }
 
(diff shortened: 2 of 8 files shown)


src/frontend/menu.scala
```diff
@@ -167,7 +167,8 @@ object FuryMenu {
       ),
       Menu('imports, msg"view and modify layer imports", 'list, shortcut = 'i')(
         Action('list, msg"show all layer imports", UniverseCli(_).imports.list, shortcut = 'l'),
-        Action('pull, msg"update layer imports universally", UniverseCli(_).imports.pull, shortcut = 'p')
+        Action('pull, msg"update layer imports universally", UniverseCli(_).imports.pull, shortcut = 'p'),
+        Action('replace, msg"replace all entries of a layer import with another layer", UniverseCli(_).imports.replace, shortcut = 'r')
```
propensive: Could we combine these and just have one command, `update`?

odisseus: Done

src/repo/universe.scala
```diff
@@ -96,30 +96,37 @@ case class UniverseCli(cli: Cli)(implicit val log: Log) extends CliApi {
...
   }
 }
 
 case class UniverseApi(hierarchy: Hierarchy) {
 
+  private[this] lazy val universe: Try[Universe] = hierarchy.universe
+
+  private[this] def findLayer(ref: ShortLayerRef): Try[LayerEntity] = universe >> (_.imports) >>= (find(_, ref))
+
+  //TODO move this to Resolver or something
```
propensive: Agreed. I think that if we have an implicit `Resolver` set up there, then we can write `repoSets.findBy(repoSetId)` below.

src/repo/universe.scala
```diff
@@ -133,38 +140,38 @@ case class UniverseApi(hierarchy: Hierarchy) {
 
   object projects {
     def proliferate(pointer: Pointer, projectRef: ProjectRef)(implicit log: Log): Try[Hierarchy] = for {
+      // FIXME looks like a bug; layer isn't used anywhere
```
propensive: Agreed, we can remove it.

src/repo/universe.scala
```diff
@@ -133,38 +140,38 @@ case class UniverseApi(hierarchy: Hierarchy) {
...
       layer     <- hierarchy(pointer)
-      universe  <- hierarchy.universe
+      universe  <- universe
```
propensive, line 139: Was there a reason for this change? I remember that I added `universe` to the `Hierarchy` type because I needed it for some reason, and I think, because it's different from the "global" universe. I don't remember the whole motivation, but I didn't really want to hold a reference to `universe` with the Hierarchy, but seem to recall that I had no option...

src/repo/universe.scala
```diff
@@ -133,38 +140,38 @@ case class UniverseApi(hierarchy: Hierarchy) {
...
                        val newImport = imp.copy(layerRef = newLayerRef, remote = pub)
                        Layer(_.imports).modify(layer)(_ - imp + newImport)
                      }
     } yield hierarchy
+
+    private def getRemoteName(layerRef: ShortLayerRef, layerEntity: LayerEntity): Try[LayerName] = for {
+      layer  <- hierarchy(layerEntity.imports.head._1)
+      imp    <- layer.imports.findBy(layerRef)
+      remote <- imp.remote.ascribe(NoPublishedName(layerRef))
+    } yield remote.url
```
propensive, line 169: Would it be possible to move this into the `Layer` type?

propensive (approved):
Looks good, and the code seems cleaner, too. Ready to merge, but you might like to do some of the tweaks I suggested first, if they're not too annoying.

src/repo/universe.scala
odisseus: The universe is used six times in the methods of `UniverseApi`, so I decided to store it in a lazy val for convenience. Inside `Hierarchy`, the universe already is a lazy val, so this is just a cosmetic change. If you don't like it, i can revert that change easily.

Since the `UniverseApi` is a case class with a hierarchy as its field, I assume that, in this context, the universe will be always the one that comes from that hierarchy.

src/repo/universe.scala
odisseus: Done. See #1515.

src/repo/universe.scala
odisseus: Done in #1514.

src/repo/universe.scala
propensive: I think it's fine. I think I've also described the earlier situation incorrectly: I meant to say that `Universe` has a reference to the `Hierarchy` that created it. I think that's generally undesirable in the long term, but it's irrelevant in the context I've brought it up here. So let's keep your change.

propensive:
Are you happy to make it a non-draft PR yet?

src/repo/universe.scala
odisseus: I'll try to do that in another pull request.

PR #1513, opened 2020-08-14
Skip the information about JAVA_TOOL_OPTIONS

odisseus:
This looks like an old problem that resurfaced after a recent update of OpenJDK 8. Hopefully we'll be able to revert this commit once the problem is fixed in the Github runner environment.


propensive (approved):
It's a bit hacky, but it's ok for now.

PR #1514, opened 2020-08-15
Removed --layer parameter from the projects proliferate command.

odisseus:
This parameter doesn't seem to affect anything.


propensive (approved):
On the basis that it's not used, we should remove it. The intention was that it would allow proliferation of a project just within one imported layer, not globally. For consistency with other commands, we should keep it... but only if it works.

PR #1518, opened 2020-08-17
Cache the makefile output for repeated builds of the same commit

odisseus:
I hope this will shave off about three minutes from each re-build.


propensive (commented):
That would be a good improvement. I generally dislike that what started as a simple build configuration file is starting to look like a script, but, here we are. ;)

Issue #1291, opened 2020-04-21
`fury layer pull` doesn't save changes


propensive:
`fury layer pull` goes through the motions of checking the remote service for updates, but doesn't seem to save them.

propensive:
@odisseus Could you take a look at this issue? It's tangentially related to what you were looking at with the universal layer updates. If it doesn't work in all cases (which I suspect to be the case) then it could benefit from a complete rewrite, which would probably be easier than trying to fix the existing logic.

PR #1520, opened 2020-08-18
Select version when importing a layer

odisseus:
Outstanding problems:
* The `--version` argument is allowed (and ignored) even when the import is referenced by its hash
* Some import hashes aren't present in the catalog
  * Sometimes migrating to a version present in the catalog yields an unexpected hash in the import list
  * This might be caused by migration of older layer versions
* The current `platform` layer contains two different references to `gastronomy`
  * Nevertheless, Fury is being built without conflict
  * Perhaps these two layers become identical after migration?
* It is impossible to get the full layer hash by its `ShortLayerRef`

This pull request s related to the issue #1291.

### DIFF
--- a/src/model/ids.scala
+++ b/src/model/ids.scala
@@ -333,7 +333,7 @@ case class ShortLayerRef(key: String) extends Key(msg"layer")
 
 case class LayerProvenance(ref: ShortLayerRef, imports: Map[Pointer, Import]) {
   def +(newImports: Map[Pointer, Import]) = copy(imports = imports ++ newImports)
-  def ids: Set[ImportId] = imports.values.map(_.id).to[Set]
+  def ids: SortedSet[ImportId] = SortedSet[ImportId]() ++ imports.values.map(_.id)
   def published: Set[PublishedLayer] = imports.values.flatMap(_.remote).to[Set]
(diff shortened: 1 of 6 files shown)


propensive (approved):

propensive:
Later, we can add some completions for layer versions.

src/model/ids.scala
```diff
@@ -333,7 +333,7 @@ case class ShortLayerRef(key: String) extends Key(msg"layer")
 
 case class LayerProvenance(ref: ShortLayerRef, imports: Map[Pointer, Import]) {
   def +(newImports: Map[Pointer, Import]) = copy(imports = imports ++ newImports)
-  def ids: Set[ImportId] = imports.values.map(_.id).to[Set]
+  def ids: SortedSet[ImportId] = SortedSet[ImportId]() ++ imports.values.map(_.id)
```
propensive, line 336: Does `.to[SortedSet]` not work?

propensive (commented):
Looks good, thank you.

src/model/ids.scala
odisseus: Looks like it could work. Will fix later.

PR #1523, opened 2020-08-21
File/directory completions partially working

### LINKED ISSUE #1516: no tab-completions are offered on `--path`/`-f` parameters
They should be. It might be possible to do this by returning something like `_files` as a single tab completion in order to trigger shell-level file completions.

propensive:


propensive:
This needs some improvements, but they can be done later.

PR #1525, opened 2020-08-21
Detection of repo conflicts

### LINKED ISSUE #1521: project conflicts don't take into account repo differences
Two projects with identical repo references but different repos will be considered "the same", which is a serious bug.

odisseus:
Fixes #1521 (I hope).

### DIFF
--- a/src/core/layer.scala
+++ b/src/core/layer.scala
@@ -56,7 +56,7 @@ case class Layer(version: Int,
 
   def localUniverse(hierarchy: Hierarchy, path: Pointer): Universe = Universe(
     hierarchy = hierarchy,
-    projects = projects.map { project => project.id -> Uniqueness.Unique(project.projectRef, Set(path)) }.toMap,
+    projects = projects.to[List].map { project => project.id -> Uniqueness.Unique(project.projectRef(repos), Set(path)) }.toMap,
     repoSets = repos.groupBy(_.commit.repoSetId).mapValues(_.map(_.ref(path))),
     imports = imports.map { i => i.layerRef.short -> LayerProvenance(i.layerRef.short, Map(path -> i)) }.toMap
   )
--- a/src/core/project.scala
+++ b/src/core/project.scala
@@ -43,13 +43,18 @@ case class Project(id: ProjectId,
                    description: String = "",
                    compiler: Option[CompilerRef] = None) {
 
-  def projectRef: ProjectRef = {
-    val digest = (id, description, main, license, modules.map { m =>
-      (m.id, m.kind, m.manifest, m.compiler, m.dependencies.to[List], m.opts.to[List], m.binaries.to[List],
-          m.environment.to[List], m.policy.to[List], m.hidden, m.optDefs.to[List], m.deterministic,
-          (m.sources.to[List] ++ m.resources.to[List])) // FIXME: Resolve these in the hierarchy!
-    }.to[List]).digest[Sha256]
+  def projectRef(repos: SortedSet[Repo]): ProjectRef = {
+    val modulesInput = modules.to[List].map(digesterInput(repos, _))
+    val digest = (id, description, main, license, modulesInput).digest[Sha256]
     ProjectRef(id, Some(digest.encoded[Hex].take(6).toLowerCase))
+  }
+
+  private[this] def digesterInput(repos: SortedSet[Repo], m: Module) = {
+    val inputs = (m.externalSources ++ m.externalResources)
+    val rs = repos.filter{ r => inputs.map(_.repoId).contains(r.id) }
+    (m.id, m.kind, m.manifest, m.compiler, m.dependencies.to[List], m.opts.to[List], m.binaries.to[List],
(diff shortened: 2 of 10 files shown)


src/core/layer.scala
```diff
@@ -54,9 +54,11 @@ case class Layer(version: Int,
     })
   })
 
-  def localUniverse(hierarchy: Hierarchy, path: Pointer): Universe = Universe(
+  def localUniverse(hierarchy: Hierarchy, path: Pointer): Try[Universe] = for {
+    foobar <- projects.to[List].traverse { project => project.projectRef(hierarchy) >> (Uniqueness.Unique(_, Set(path))) >> (project.id -> _) }
+  } yield Universe(
     hierarchy = hierarchy,
-    projects = projects.map { project => project.id -> Uniqueness.Unique(project.projectRef, Set(path)) }.toMap,
+    projects = foobar.toMap,
```
propensive: We still need better names than `foobar`...

src/core/project.scala
```diff
@@ -43,15 +43,24 @@ case class Project(id: ProjectId,
...
   }
 
+  private[this] def digesterInput(hierarchy: Hierarchy, m: Module) = {
+    for {
+      universe <- hierarchy.universe
+      rs <- universe.repoSources(m.ref(this))
+    } yield {
+      (m.id, m.kind, m.manifest, m.compiler, m.dependencies.to[List], m.opts.to[List], m.binaries.to[List],
+        m.environment.to[List], m.policy.to[List], m.hidden, m.optDefs.to[List], m.deterministic,
+        (m.sources.to[List] ++ m.resources.to[List]), rs) // FIXME: Resolve these in the hierarchy!
```
propensive: I think this is ok. I've read it a few times.

src/core/universe.scala
```diff
@@ -83,16 +83,19 @@ case class Universe(hierarchy: Hierarchy,
     case (acc, (_, Ambiguous(origins))) => acc ++ origins.values
   }
 
-  def checkout(ref: ModuleRef, hierarchy: Hierarchy, layout: Layout)(implicit log: Log): Try[Snapshots] = for {
+  def repoSources(ref: ModuleRef): Try[List[(Repo, List[RepoSource])]] = for {
```
propensive: The return type suggests the order is significant. Shouldn't this be a `Map`?

test/pending/commits-conflict/script
```diff
@@ -0,0 +1,17 @@
...
+fury universe projects
+
+fury layer select -l /magnolia
+fury repo list --raw --column Commit
+fury repo update -r magnolia -t v0.17.0
+fury repo list --raw --column Commit
+
+fury layer select -l /
+fury universe imports list #--raw --column Import
+fury universe projects
```
propensive: Sholudn't this line produce some output? I can't see anything in the `check` file.

odisseus:
I accidentally a circular dependency between `ProjectRef` and `Universe`. Now I have to ditch everything and start over.

Basically, you cannot get the state of the repos while constructing the project, because this requires the layer to be already constructed, including the current project.

This also means that you cannot check whether all the project have consistent requirements to the state of repos in the universe until the universe is completely constructed.

test/pending/commits-conflict/script
odisseus: The test isn't complete yet.

propensive:
I'm not sure it should be an issue. Repos are scoped to the current layer, and we don't need a `Universe` to construct each project, in the context of its layer, but the method which constructs the universe probably needs to construct the `RepoSet`s first, and then pass these as a parameter to the line which constructs the `ProjectEntity`s.

src/core/layer.scala
odisseus: Refactored

src/core/project.scala
odisseus: I just copied this tuple as it was, and added a new element at the end.

src/core/universe.scala
odisseus: Refactored

test/pending/commits-conflict/script
odisseus: Fixed

test/passing/projects-proliferate/script
```diff
@@ -41,6 +41,6 @@ fury layer import -l $IPFSREF2 --name platform2
 echo $?
 fury layer list --raw --column ref
 fury universe projects list --raw
-fury universe projects proliferate -p hello-world@8803c7
+fury universe projects proliferate -p hello-world@9a328e
 fury layer list --raw --column ref
 fury universe projects list --raw
```
propensive, line 46: Shouldn't this line list the `hello-world` project, without the `@hash`?

propensive (approved):

test/passing/projects-proliferate/script
odisseus: It does. The command at line 46 produces the last three lines in the check file:
```
U2Asaiwr
hello-world
scala
```

PR #1527, opened 2020-08-22
Always output ANSI reset at the end of output

### LINKED ISSUE #1526: always reset terminal's ANSI state after output
Certain commands can leave the terminal displaying output in "bold". This can (hopefully) be fixed by always sending an ANSI "reset" after all output.

propensive:


propensive:
This didn't work as expected... the integration tests fail quite universally...

PR #1532, opened 2020-08-24
Better error when `java` or `javac` missing

propensive:


propensive:
Includes other improvements to the installation, too.

PR #1538, opened 2020-08-25
Use vent.dev instead of furore.dev

propensive:


odisseus:
Why?

> Vent, the cloaca region of an animal

https://en.wikipedia.org/wiki/Vent

propensive:
Venting is what one does with ones fury.

PR #1542, opened 2020-08-25
Removed the hierarchy argument from the methods that don't use it.

odisseus:


propensive (approved):
Looks good in principle.

PR #1553, opened 2020-08-27
Install Java if necessary

propensive:

PR #1555, opened 2020-08-29
Improved CI configuration

odisseus:
* Split the CI checks into three phases: compilation, unit tests, and integration tests
* Trigger CI only on pull request events
* Bumped actions/upload-artifact to version 2
* Use openssl to calculate MD5 hashes in the scripts
* Use readlink and realpath to find Fury's actual installation location if both commands are available
* Don't assume "system install" task if the output of the launcher script is consumed by a pipe
* Changed the structure of installation directories (same as in PR #1553)
* Changed the format of log archives

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,16 +1,14 @@
 name: Build
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
-
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
       fail-fast: false
-
     steps:
     - uses: actions/checkout@v2.0.0
     - run: git fetch --prune --unshallow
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -22,54 +20,101 @@ jobs:
           dist
           etc/launcher
         key: dist-${{ github.sha }}
+    - name: Install IPFS (Mac OS)
+      if: startsWith(matrix.os, 'macos')
+      run: brew install ipfs
+    - name: Install IPFS (Ubuntu)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: sudo snap install ipfs
```
propensive, line 28: Having these here means that we don't get to test the (common) path where most people will not have IPFS installed.

.github/workflows/main.yml
```diff
@@ -22,54 +20,101 @@ jobs:
...
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v2.0.0
+    - uses: actions/download-artifact@v2
+      with:
+        name: build-dist-${{ runner.os }}
+        path: dist
+    - name: Install Fury
+      run: |
+        chmod +x dist/fury
```
propensive, line 62: Shouldn't this happen in the `Makefile`?

Makefile
```diff
@@ -113,7 +113,7 @@ dist/fury: etc/launcher tmp/.bundle.ipfs
 	 mkdir -p dist && \
 	 sed -e "s/%VERSION%/$(VERSION)/" \
 	     -e "s/%HASH%/$(shell cat tmp/.bundle.ipfs)/" \
-	     -e "s/%MD5%/$(shell md5sum dist/fury.tar.gz | head -c 32)/" "$<" > "$@" && \
+	     -e "s/%MD5%/$(shell openssl dgst -md5 dist/fury.tar.gz | grep -E -o '[a-f0-9]{32}')/" "$<" > "$@" && \
```
propensive: Is there a reason for using `openssl` instead of `md5sum`?

etc/launcher
```diff
@@ -53,9 +53,8 @@ download() {
 }
 
 checkFile() {
-  command -v md5sum > /dev/null || command -v md5 > /dev/null || fail md5sum
   printf "${info}Checking MD5 checksum of ${downloadFile}..."
-  calcHash=$( (md5sum "${downloadFile}" 2> /dev/null || md5 -r "${downloadFile}" 2> /dev/null) | head -c 32)
+  calcHash=$(openssl dgst -md5 "${downloadFile}" | grep -E -o '[a-f0-9]{32}')
```
propensive: We can't assume that `openssl` is installed in the script's environment, unfortunately.

test/passing/import-update/check
```diff
@@ -16,7 +16,12 @@ Set current module to app
 Checking out src/core from repository magnolia
 Cloning repository at gh:propensive/probably
 Checking out src/cli from repository probably
+Cloning repository at gh:propensive/mercator
 Checking out src/core from repository mercator
+Cloning repository at gh:propensive/escritoire
+Checking out src/core from repository escritoire
+Cloning repository at gh:propensive/gastronomy
+Checking out src/core from repository gastronomy
```
propensive, line 24: Are these new lines here because we didn't just build Fury in the same environment, and hence it has become necessary to include them?

propensive (changes requested):
All looks great, apart from using `openssl` in the launcher.

Makefile
odisseus: It is present in both Linux and Mac OS environments.

test/passing/import-update/check
odisseus: Yes. These repositories used to be checked out during the initial compilation or the unit tests. But now the integration tests run in a clean environment, so it is necessary to check them out once again.

etc/launcher
odisseus: If you mean Windows, I believe it doesn't have `md5` or `md5sum` either.

.github/workflows/main.yml
odisseus: This affects only the environment in which the makefile is executed, and the makefile already assumes that IPFS is installed.

The tests are run without pre-installed IPFS.

.github/workflows/main.yml
odisseus: The makefile does this too, but apparently uploading and downloading the artifacts doesn't preserve file permissions.

etc/launcher
propensive: It's available on all Linux distributions, Mac OS X, and Cygwin (which doesn't work for other reasons anyway). OpenSSL is must less commonly available.

propensive: It's not part of a standard Debian or Ubuntu server install, for example.

propensive (changes requested):

propensive:
@odisseus Try something like this:
```
(md5sum $file 2> /dev/null || md5 -r $file 2> /dev/null) | head -c 32
```

etc/launcher
odisseus: Refactored

PR #1558, opened 2020-09-02
Try to avoid Java 8/11 mismatch issue

### LINKED ISSUE #1557: Fury distribution built with Java 11 is incompatible with Java 8
## Steps to reproduce:

1. Set the default JVM to Java 11.
1. Run `make install`.
1. Make sure Fury and Bloop are not running.
1. Set the default JVM to Java 8.
1. Run `fury standalone`.

The cause of incompatibility is explained [on Stackoverflow](https://stackoverflow.com/questions/61267495/exception-in-thread-main-java-lang-nosuchmethoderror-java-nio-bytebuffer-flip).

## Stack trace:
```
Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
	at fury.ogdl.Ogdl$.$anonfun$readToBuffer$1(ogdl.scala:156)
	at scala.util.Try$.apply(Try.scala:213)
	at fury.ogdl.Ogdl$.readToBuffer(ogdl.scala:148)
	at fury.ogdl.Ogdl$.read(ogdl.scala:139)
	at fury.core.Layer$.$anonfun$readFuryConf$2(layer.scala:290)
	at scala.util.Success.flatMap(Try.scala:251)
	at fury.core.Layer$.$anonfun$readFuryConf$1(layer.scala:289)
	at scala.util.Success.flatMap(Try.scala:251)
	at fury.core.Layer$.readFuryConf(layer.scala:286)
	at fury.FuryServer$.$anonfun$invoke$1(main.scala:39)
	at scala.util.Success.flatMap(Try.scala:251)
	at fury.FuryServer$.invoke(main.scala:38)
	at fury.FuryServer$.$anonfun$run$2(main.scala:83)
	at fury.core.Lifecycle$.trackThread(lifecycle.scala:80)
	at fury.FuryServer$.run(main.scala:82)
	at fury.FuryServer$.main(main.scala:59)
	at fury.FuryServer.main(main.scala)

```

propensive:


odisseus:
Unfortunately, this doesn't fix the issue.

odisseus (changes requested):
Unfortunately, this doesn't fix the issue.

propensive:
I wonder if Scala considers the `match` to be a no-op rather than a cast because it's _statically_ true already. I'll try casting it to an object first to see if that makes a difference.

PR #1566, opened 2020-09-07
Increase timeout for integration tests to 20 minutes

odisseus:


propensive (approved):
I think this isn't the solution, but accepting it on the basis that it's useful anyway.

PR #1572, opened 2020-10-30
Update tutorial.md

nathanknox:
`scala-new` isn't an option currently in the list of layers. The options available that make the most sense to me are `scala3-new` and `scala`. I went with vanilla `scala` - if this is incorrect feel free to reject!


nathanknox:
scala is wrong. looks like scala-new is missing. Playing with scala3-new in the meanwhile.

PR #1627, opened 2020-12-24
fix backticks on concepts.md

ivanopagano:


propensive (approved):

propensive:
Thank you, Ivano! I'm also planning an overhaul of that file over the next week... Buon Natale!

